PR #54, opened 2014-08-15
Keep paths in main bundle names to ensure unique bundle names, fixes #52

### LINKED ISSUE #52: Names generated for main bundles may override one another
When building an app with several main modules, the main module's filename is used as the name for its bundle filename. Since all bundles are put into one folder, it's possible for one to override another.

Take a hypothetical app that builds two main modules, `users/edit/edit` and `tasks/edit/edit`:

1. `tasks/edit/edit.js` is built into bundle `dist/bundles/edit.js`
2. `users/edit/edit.js` is built into bundle `dist/bundles/edit.js`, overriding the first bundle.

Multi-build in legacy-steal avoided this by naming the bundle `production.js` and putting in the same directory as the originating bundle. (But I do prefer this new behavior of keeping all production files in one `dist` folder, though.)

Some potential solutions:

1. Use the relative path of the main bundle, replacing each "/" with "__", for the name of the bundle. That is, `tasks/edit/edit.js` would become `dist/bundles/tasks__edit__edit.js`. This is what I have hacked into my test project.
2. Recreate the paths in the `dist/bundles` folder. That is, `tasks/edit/edit.js` would become `dist/bundles/tasks/edit/edit.js`.

Important to note that I believe StealJS would need to be updated to do the same thing in production mode, too, so whatever naming scheme is used, it must be possible for a client to do also. That means we can't use incrementing numbers or such.

asavoy:
Fixes #52.

Main bundle names should use the full path. If we only use the filename, we may conflict with other main bundles. 

Example: `users/list.js` and `tasks/list.js` would both be bundled to `bundles/list.js`; whichever is built last would overwrite the first.

With this fix, `users/list.js` would build to `bundles/users/list.js`, and `tasks/list.js` would build to `bundles/tasks/list.js`.

There may be other schemes to generate a bundle name but they a) are more complicated and/or b) can't guarantee uniqueness as well as just using the pathname.

**NOTE:** This also requires a matching change to StealJS' `steal.js` and `steal.production.js`. In `addProductionBundles()`:

    mainBundleName = bundlesDir+filename(main);

should become:

    mainBundleName = bundlesDir+main;

I'll submit a PR for that too. **EDIT:** Added a patch to steal here: https://github.com/bitovi/steal/pull/222

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "steal-tool",
   "version": "0.1.0",
   "devDependencies": {
-    "steal": "git://github.com/bitovi/steal#a6e7eda6cd99b96eb9e18ae4f6ccf6c80156dfea",
+    "steal": "git://github.com/asavoy/steal#771ee6840c396a91a5dbc1ae266c9648c6882d14",
     "jquery": "~1.10.0",
 	"less" : "^1.7.0"
   }
--- a/lib/bundle/name.js
+++ b/lib/bundle/name.js
@@ -13,11 +13,15 @@ function getName(bundle) {
 		return (appName+"").replace(".js","");
 	});
 
-	// try with just the last part
 	var shortened = bundleNames.map(function(l){
 		return filename(l);
 	}).join('-');
 
+	// If this is a "main" bundle
+	if (bundle.bundles.length === 1 ) {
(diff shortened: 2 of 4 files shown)


alexisabril:
Closed by PR #157

PR #103, opened 2014-10-13
Fix path separator for urls on windows

whitecolor:


matthewp:
This is done in https://github.com/bitovi/steal-tools/commit/9475fe1354c10ec53620baeb57308416c6c80495

Issue #145, opened 2015-01-10
Bundle paths generated on Windows builds gets corrupted


booc0mtaco:
Noticed this when inspecting the resulting build file after using `steal-tools` with grunt.

Given the following configuration:

```
options: {
	system: {
		config: __dirname + '/stealconfig.js',
		main: 'test/test',
		bundlesPath: './path/build/production/'
	},
}
```

a built file is produced at the correct path, but has the following for `System.paths`

```
System.paths["bundles/*.css"] ="path\build\production/*css";
System.paths["bundles/*"] = "path\build\production/*.js";
```

Which causes problems on Windows. I was able to patch this in `steal-tools` by editing 
https://github.com/bitovi/steal-tools/blob/master/lib/bundle/make_bundles_config.js#L43

and inserting `.replace(/\\/g, "/")` but I'm not sure if this is the best solution, or the proper place. 

Any ideas?

justinbmeyer:
somehow the bundlesPath is changing from / to \?  I wonder why.

booc0mtaco:
One of the path libraries must be outputting a Windows-formatted directory string. Since that's being inserted into a new JS file, it gets interpreted as escape sequences when the built file is parsed. 

https://github.com/bitovi/steal-tools/blob/master/lib/bundle/make_bundles_config.js#L46

That library *might* have options to change the output format. `/` works perfectly fine on both platforms. I haven't had a chance yet to find out where that's happening, or if the library has such an option.

booc0mtaco:
After reviewing the [path](http://nodejs.org/docs/latest/api/path.html#path_path_relative_from_to) library, found that the `.relative` method outputs platform-formatted strings. This path value gets created in 

<https://github.com/bitovi/steal-tools/blob/master/lib/configuration/make.js#L40>

Which is fine for all file operations except when creating the output build files. Here, the path should be normalized for the browser and use forward slashes. Since the property is called `bundlesPathURL`, I *think* the `.replace(/\\/g, '/')` should be in make.js above instead of in 

https://github.com/bitovi/steal-tools/blob/master/lib/bundle/make_bundles_config.js#L43

Thoughts? @matthewp @justinbmeyer

whitecolor:
An old issue.

justinbmeyer:
@whitecolor Do you mean this is no longer a problem and can be closed?

whitecolor:
I mean that it wan't fixed too long.

matthewp:
[...]

PR #147, opened 2015-02-03
RE #145 Add `.replace` call to normalize output of `bundlesPathURL` with...

booc0mtaco:
... forward slashes on all platforms.


justinbmeyer:
Thanks!  Do you have a windows machine?  Do all the tests break and this fixes them?  I'm wondering if I need a test.

PR #158, opened 2015-02-16
Transpiles with 6to5

matthewp:
This adds support for using 6to5 as part of the build. Since this is defined in the build most of the work is done by transpile. All we have to do here is make sure the option which is on the loader is passed as a transpile option.


matthewp:
Updating the name of the option that gets passed to transpile, will reopen once that's done.

Issue #89, opened 2014-10-01
stealBuild option system.main issue


whitecolor:
In dev mode main (via script tag attribute) can be defined as for example ```apps/first/``` that will load ```apps/first/first.js```.

In stealBuild options if system.main is defined as ```apps/first/``` will result in build error, so one need to define it there as ```apps/first/first```. 

A little bit inconsistent in my view. Not sure if its needed and correct behaviour.

justinbmeyer:
@joe-crick Once you get the docs working, maybe we can tackle this together.

m-mujica:
I'll take this one. :cop:

matthewp:
I need this for https://github.com/bitovi/steal-tools/issues/168 so I'm taking it.

m-mujica:
![tumblr_m9s72v0ukt1ro2xth](https://cloud.githubusercontent.com/assets/724877/6381892/66c76d0c-bd23-11e4-95e6-e36b2b13a891.gif)

matthewp:
Closed by #169

PR #169, opened 2015-02-26
Normalize main and bundles

### LINKED ISSUE #168: Make work with new directories.lib
The new directories.lib changes the way modules can be imported from inside the package so need to update tests.

Partially fixed in system-npm: https://github.com/bitovi/system-npm/pull/12

matthewp:
This updates the build process to normalize System.main in dependencyGraph so that users can pass unnormalized mains as the `main` option of `system`.

Also updates tests to work with the new directories.lib. Closes #168 and #89

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "steal-tools",
   "version": "0.5.0",
   "devDependencies": {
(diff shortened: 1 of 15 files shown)


matthewp:
Once this passes will merge the dependent PRs upstream (steal and system-npm) and then will update this before merging.

PR #190, opened 2015-03-13
Support export with module.exports

### LINKED ISSUE #189: global shim should test for "module"
The Babel export for amd creates something like;

```js
define([
  "exports",
  "module"
], function(exports, module) {
  module.exports = "foo";
});
```

We are currently handling `exports` case but not `module`.

matthewp:
Babel spits out AMD code that makes use of an `exports` and `module` object.  This is different from other cjs->amd transpilation with uses `exports`, `module`, and `require`. Since Babel doesn't use `require` we need to shim this behavior.  It's as simple as checking the names of the arguments and that there is not an existing import for them. Fixes #189


matthewp:
This test is going to fail because it is blocked by https://github.com/bitovi/transpile/issues/20

matthewp:
Going to hold off on this one.

PR #226, opened 2015-04-17
Watch mode

### LINKED ISSUE #80: Watch mode
It would be nice if we could produce a "watch mode". This would be part of the build and pluginify. It would watch for changes in the files in your graph and reload just those files (and their dependencies). It would be a very fast build.

Along with this we should also support `node_modules` just for watch mode (not in the browser). This would be very simple to implement, it would just be a `fetch` hook that, if the fetch fails, uses `require.resolve` to look-up the path in node_modules, if present, and then retry with that path.

These 2 things would give us 90% of the things people like about Browserify and WebPack; near-instant builds in development mode and using npm so it "just works".

matthewp:
The watch mode is a mode of the multi-build that continously builds as files in the dependency graph change. To use:

```shell
steal-tools --watch
```

This will use `package.json!npm` as the config (you can still pass `--config` and `--main` options if you want), do an intial build and watch for file changes which will trigger a rebuild.

Closes #80

### DIFF
--- a/bin/steal
+++ b/bin/steal
@@ -15,7 +15,10 @@ var yargs = require("yargs")
 	.alias("o", "out")
 	.describe("source-maps", "Generate source maps")
 	.describe("source-maps-content", "Include the original source contents with the source maps")
+	.describe("minify", "Minify the output. Defaults to true except when used with --watch")
 	.describe("no-minify", "Do not minify the output")
+	.describe("w", "Watch for file changes and rebuild")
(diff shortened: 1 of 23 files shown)


matthewp:
Working on docs now.

marshallswain:
Woot!  Doing a happy dance!  Maybe this will buy me some more time with my trusty 2009 Macbook Pro.

PR #232, opened 2015-04-22
Making bundleSteal option work for bundling a webworker with steal

### LINKED ISSUE #231: BundleSteal in a web worker fails with `window is not defined`
I want to be able to build/bundle a web worker file with Steal.

When I bundle the web worker with steal (`bundleSteal: true`) and I try and use the web worker it fails with the error:

```javascript
Uncaught ReferenceError: window is not defined
```

The offending code seems to be here:
https://github.com/stealjs/steal-tools/blob/master/lib/bundle/add_steal.js#L8

in the `[add-define]` part.

Simply changing it to `this.define` instead of `window.define` seems to work, but I am not sure of the consequences of such a change.

I could make a PR but I need some help figuring out how to write a test for this.

Here is a gist with some example code if you like:  
https://gist.github.com/BigAB/89c86d54c919d04a23d1

You'll need to `npm install`, `bower install` and `grunt` then serve `index.html` and `build.html` for the build version.

BigAB:
Fixes #231 

Replaces the line:

```javascript
window.define = System.amdDefine;
```

with
```javascript
((typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope) ? self : window).define = System.amdDefine;
```

...so that it can run in a web worker.

Adds a test to ensure it works. But not sure how to hook up the test to CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ node_js:
 before_script:
   - npm install -g bower
   - bower install
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,24 +4,50 @@ module.exports = function(grunt) {
 
   // Project configuration.
   grunt.initConfig({
-    simplemocha: {
-      app: {
-        src: ['test/test.js']
-      }
-    },
-    jshint: {
-      options: {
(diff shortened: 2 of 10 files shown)


matthewp:
Thanks, i'll wire it up to CI.  Also I think i'm going to move the build stuff into the Gruntfile.js

That way we are testing our Grunt tasks for free.

PR #245, opened 2015-04-30
Normalize modelNames only if typeof === string

lukewhyte:
Could be string, function or RegEx


stevenvachon:
What about `instanceof String` which would be `typeof==="object"`?

lukewhyte:
I've expanded the conditional to check for instanceof String too.

matthewp:
Thanks.

PR #250, opened 2015-05-18
sideBundle: true to create a side bundle for a module

### LINKED ISSUE #14: bundle: false
We should be able to tell a module that it should not be packaged in production.  We should explore how requirejs does this. If requirejs doesn't have an API for it, I suggest putting it on meta:

System.meta.html5Shiv = {bundle: false};

matthewp:
This adds the ability to create a side bundle for a module.

```js
System.config({
  meta: {
    "some/module": {
      sideBundle: true
    }
  }
});
```


matthewp:
Going to reopen against minor.

PR #268, opened 2015-06-30
Make Grunt task fail when build fails.

### LINKED ISSUE #263: Grunt task should always fail when there's an error
Currently the [build Grunt task](https://github.com/stealjs/steal-tools/blob/master/tasks/build.js) does not error when the build fails.

m-mujica:
Closes #263.

high-


matthewp:
Looks good, thanks.

Issue #205, opened 2015-04-01
Document noGlobalShim


matthewp:
Used [here](https://github.com/stealjs/steal-tools/blob/master/lib/build/transform.js#L104) and should appear on [this page](http://stealjs.com/docs/steal-tools.transform.options.html)

ilyavf:
Used [here](https://github.com/stealjs/steal-tools/blob/master/lib/build/transform.js#L123)

chasenlehara:
@ilyavf, do you have commits for this issue that should be merged in?

PR #270, opened 2015-07-07
Added documentation to noGlobalShim option #205

ilyavf:


doc/types/transform-options.md
```diff
@@ -33,6 +33,9 @@ name on the global object in exports.
 
 The other possible format values are "steal","amd", and "cjs".
 
+@option {Boolean} [noGlobalShim=false] By default, if _options.format_ is `"global"` then a shim will be added
+for the global _define_ method. Setting _noGlobalShim_ to `true` prevents this.
+
```
matthewp: This is good, perhaps also explain that the global shim is so that the script can be ran standalone, and exluding the shim means it will have to be ran with an AMD loader like requirejs.

matthewp:
Thanks.

PR #281, opened 2015-08-03
Pass in all bundleAsset options into steal-bundler

matthewp:


matthewp:
Need a new pre-release for this fix.

PR #285, opened 2015-08-07
Correctly name bundles that use directories.lib

### LINKED ISSUE #284: Build not naming bundles correctly when using directories.lib
I'm seeing a main bundle like `dguide@0.js` which is wrong.

matthewp:
Because the module name for a main will be something like:

```
package@1.0.0#module/path
```

You will need to load this in production with a script tag like:

```html
<script src="steal.production.js"
  main="package/module/path"></script>
```

And therefore the bundle name should be `bundles/package/module/path`.

Fixes #284

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "steal-tools",
   "version": "0.11.0-pre.6",
   "devDependencies": {
-    "steal": "^0.11.0-pre.0",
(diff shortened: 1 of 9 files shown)


matthewp:
This needs https://github.com/stealjs/system-npm/pull/70

PR #316, opened 2015-10-02
fixing IE8 "Expected identifier" Error when injecting import statement

smoothlikejazz:
Hi Mathew, 

Just doing some testing in IE and found dot notation in add_steal. So I updated the dot notation to use square bracket notation, otherwise IE8 complains about Expected identifier.
screen shot 2015-10-02 at 4 23 10 pm


matthewp:
Thanks! Will get this in a patch release Monday.

smoothlikejazz:
No problem at all.

Issue #319, opened 2015-10-08
Pass babelOptions and traceurOptions to transpile


matthewp:
Branch from minor for this.

matthewp:
We have this for babel. I don't think there's much interest in Traceur any more. Happy to accept a PR for this.

PR #321, opened 2015-10-08
adding test for traceurOptions

phillipskevin:
Test for `traceurOptions` to support https://github.com/stealjs/steal-tools/issues/319 which is handled by https://github.com/stealjs/transpile/pull/45.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -412,13 +412,13 @@ describe("multi build", function(){
 			multiBuild({
 				config: __dirname + "/es6-loose/config.js",
 				main: "main",
-				transpiler: "babel"
-			}, {
-				quiet: true,
-				minify: false,
+				transpiler: "babel",
 				babelOptions: {
 					loose: 'es6.modules'
 				}
+			}, {
+				quiet: true,
+				minify: false
 			}).then(function(){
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -403,6 +403,37 @@ describe("multi build", function(){
...
+				return done(error);
+			}
+
+			multiBuild({
+				config: __dirname + "/es6-loose/config.js",
+				main: "main",
+			}, {
+				quiet: true,
+				minify: false,
+				traceurOptions: {
```
matthewp: Shouldn't this be part of the first set of options?  I don't know how this is going to be on the loader otherwise.

phillipskevin: Yes, you're right. I'll make these changes. `babelOptions` should also be in the first set of options, right? I'll need to update that also.

matthewp: Yep, strange that it is working the other way.

phillipskevin: It's working because babelOptions is already on `opts` before it gets to https://github.com/stealjs/steal-tools/blob/minor/lib/graph/transpile.js#L53. That code isn't doing anything the way the test is written now.

PR #337, opened 2015-12-15
Conditionally set bundles/* paths only if not using bundlesPath

### LINKED ISSUE #334: steal bundlesPath do not work as expected
as the documentation says,

> A convience configuration property for setting the path of where the production bundles folder is located.

my build.js looks like 
```

stealTools.build({
    main: "javascript/web/index/index",
    config: __dirname + "/package.json!npm",
    bundlesPath: __dirname + "/javascript/dist"
}, {
    minify: false,
    bundleSteal: false,
    bundleAssets: false
});
```

in the bundled file the system path was set by steal-tools to:
```
System.paths["bundles/*.css"] = "../../../../javascript/dist/*css";
System.paths["bundles/*"] = "../../../../javascript/dist/*.js";
```
after the build, the css file is not loading -> wrong path

***Note: the app is located at http://localhost/mycanjsapp/***

i tried:

1. ```<script src='node_modules/steal/steal.js' env="production" bundles-path="/mycanjsapp/javascript/dist" main="javascript/web/index/index"></script>```

    **Result:** css file still not loading ```GET http://localhost/javascript/dist/javascript/web/index/index.css```

2. 
  ```
  <script>
  steal = {
   bundlesPath: "/mycanjsapp/javascript/dist",
   env: "production"
  };
  </script>
  <script src='node_modules/steal/steal.js' main="javascript/web/index/index"></script>
   ```

   **Result:** css file still not loading ```GET http://localhost/javascript/dist/javascript/web/index/index.css```

3. 
 ```
<script>
steal = {
    bundlesPath: "/mycanjsapp/javascript/dist",
    env: "production"
};
</script>

<script src='node_modules/steal/steal.js' main="javascript/web/index/index"></script>
<script>
    System.paths["bundles/*.css"] ="/mycanjsapp/javascript/dist/*css";
    System.paths["bundles/*"] = "/mycanjsapp/javascript/dist/*.js";
</script>
 ```
     **Result:** css file still not loading ```GET http://localhost/javascript/dist/javascript/web/index/index.css```

The Problem is, whatever set via ```steal``` or ```System.paths``` it will be overwritten in the bundled file, see above.

The solution would be to conditional setter in the bundled file
```
System.paths["bundles/*.css"] = (!System.paths["bundles/*.css"]) ? "../../../../javascript/dist/*css" : System.paths["bundles/*.css"];
System.paths["bundles/*"] = (!System.paths["bundles/*"]) ? "../../../../javascript/dist/*.js" : System.paths["bundles/*"];
```
in https://github.com/stealjs/steal-tools/blob/v0.13.0-pre.0/lib/bundle/make_bundles_config.js#L46-L48 somethink like that:

    var paths = util.format('System.paths["bundles/*.css"] = (!System.paths["bundles/*.css"]) ? "%s/*css" : System.paths["bundles/*.css"];\n' +
            'System.paths["bundles/*"] = (!System.paths["bundles/*"]) ? "%s/*.js" : System.paths["bundles/*"];\n',
            bundlesPath, bundlesPath);

matthewp:
If the user has set bundlesPath in the client then there is no need to set the `bundles/*` paths as those are being set in Steal. By overriding them it can cause the bundles to be loaded from the incorrect location.

Closes #334

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,3 +3,4 @@
 end_of_line = LF
 indent_style = tab
 indent_size = 4
+trim_trailing_whitespace = false
--- a/lib/bundle/make_bundles_config.js
+++ b/lib/bundle/make_bundles_config.js
@@ -1,7 +1,7 @@
 var util = require("util");
 
 /**
- * 
+ *
  * @param {Object} bundles
  * @param {Object} configuration
  * @param {Object} targetBundle the bundle this config is being built for.
@@ -10,12 +10,12 @@ var util = require("util");
 module.exports = function(bundles, configuration, targetBundle, options){
 	var paths = getBundlesPaths(configuration);
 	var excludedBundles = options.excludedBundles || {};
-	
+
 	var bundledBundles = bundles.slice(0);
 	if(configuration.options.bundleSteal){
 		bundledBundles.shift();
 	}
-	
+
 	var bundlesConfig = {};
 	bundledBundles.forEach(function(bundle){
 		// Don't write a bundles config for your own bundle.  Otherwise,
@@ -26,7 +26,7 @@ module.exports = function(bundles, configuration, targetBundle, options){
 			});
 		}
 	});
-	
+
 	return {
 		load: {name: "[system-bundles-config]"},
 		minifiedSource: paths + "System.bundles = "+JSON.stringify(bundlesConfig)+";"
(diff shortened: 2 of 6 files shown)


matthewp:
Looks like an export test is failing for unrelated reasons. Going to fix that first.

PR #344, opened 2015-12-28
Properly recycle virtual modules

### LINKED ISSUE #341: Recycling modules should detect virtual module children
live-reload and --watch mode both have a module, lib/graph/recycle that reuses (recycles) a dependency graph object when a file changes, so that rebuilds are much faster. 

One of the optimizations this has is to check if the module's dependencies have changed, and if not there is nothing that needs to be done except replace the source and remove the active source keys.

However, some types of modules should not go through this fast-path. An example is https://github.com/donejs/done-component which creates a few virtual modules. These virtual modules have their own dependencies, so merely checking if a component's dependencies have changed will give you a **false-negative**.

I think the fix I'm going to go with is that plugins can add a property to `load.metadata` that signals that the module should not be recycled. `load.metadata.unrecyclable`, most likely.  This way we'll skip the "fast path" and instead fully reload the .component modules.

matthewp:
This fixes #341

This is a good fix because it doesn't require any special configuration by plugins. Instead we detect any virtual modules that are created with System.define. When we diff a module to see if it has added/removed dependencies we run it through System.translate and System.instantiate. If translate produces virtual modules we then run the same process on the virtual modules. If anything has changed at all we know that a full trace is needed to get a fresh graph.

### DIFF
--- a/test/recycle_test.js
+++ b/test/recycle_test.js
@@ -1,5 +1,6 @@
 var assert = require("assert"),
 	bundle = require("../lib/graph/make_graph_with_bundles"),
+	fs = require("fs-extra"),
 	logging = require("../lib/logger"),
 	mockFs = require("mock-fs"),
 	path = require("path"),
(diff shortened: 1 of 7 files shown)


test/recycle_test.js
```diff
@@ -12,38 +13,6 @@ describe("Recycle", function(){
 
 	afterEach(mockFs.restore);
 
-	it("Creates an error message when there is an es6 syntax error", function(done){
```
matthewp: Nothing that this test is removed because it's no longer needed. The reason there was an error is because the modules would actually be transpiled by traceur/babel. But with [system-trace](https://github.com/stealjs/system-trace) you can prevent this with `System.preventModuleExecution = true` which I've added.

PR #348, opened 2016-01-06
Minify CSS before writing the bundles.

### LINKED ISSUE #88: Consider minifying CSS before writing the bundles
Currently we minify each css node individually. This means if there are duplicate selectors they will both make it to the final output.  We could minify again (or possibly only minify) the css prior to writing out the bundles.

m-mujica:
Closes #88.

Instead of minifying each css node individually, this PR changes steal-tools to minify the concatenated bundle before writing. I tested this branch in my current app and got some interesting numbers:

- before:
652K	dist/bundles/bundle-a.css
268K	dist/bundles/bundle-b.css
192K	dist/bundles/bundle-a-b.css

- after
440K	dist/bundles/bundle-a.css
196K	dist/bundles/bundle-b.css
192K	dist/bundles/bundle-a-b.css

There is some noise in the `write_bundles` file b/c I made some changes to the indentation, here is the interesting diff https://github.com/stealjs/steal-tools/commit/8c3da0e207bd88ba57f1b6dcc240678c6368b9ad.

### DIFF
--- a/lib/bundle/minify_source.js
+++ b/lib/bundle/minify_source.js
@@ -1,0 +1,11 @@
+var minifyCSS = require("../buildTypes/minifyCSS");
+
+module.exports = function(bundle, options) {
+	options = options || {};
+
+	if (options.minify && bundle.buildType === "css") {
+		bundle.source = minifyCSS(bundle.source, options);
+	}
+
+	return bundle;
+};
--- a/lib/bundle/write_bundles.js
+++ b/lib/bundle/write_bundles.js
@@ -3,13 +3,14 @@
 // Writes them out to the file system.
 var bundleFilename = require("./filename"),
 	concatSource = require("../bundle/concat_source"),
-	normalizeSource = require('./normalize_source'),
+	normalizeSource = require("./normalize_source"),
 	fs = require("fs"),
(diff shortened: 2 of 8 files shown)


m-mujica:
@matthewp should I send this PR against minor instead? Not sure why the build failed.

m-mujica:
The build was failing to a change in socket.io that https://github.com/socketio/socket.io/issues/2368 broke feathers/testee, a new version of testee fixed the problem.

I also noticed that my main css bundle was quite big, I was able to reduce its size by using less reference imports bringing it down to 276K minified, with this branch the same bundle minified is 272K; which is less of a dramatic improvement, but :itissomething:

matthewp:
Went over it, looks good. Trying to think of when this should go it. Is it a breaking change? Trying to think of a scenario where your app would behave differently because of the non-duplicate selectors.

matthewp:
Putting it under the 1.0.0 milestone for now, could go in sooner.

matthewp:
This will go on 0.14.0

matthewp:
This is merged into minor and part of 0.14.0-pre.0. Closing.

PR #371, opened 2016-02-14
Normalization fix

pYr0x:
fix the problem, that module names are not normalized when using package.json as config in build process

this fix some of the problems with
stealjs/steal#578


matthewp:
Can you provide a test? Might enlighten why the tests are failing. Here's an existing multiple main test: https://github.com/stealjs/steal-tools/blob/master/test/multibuild_test.js#L1081

Could use that to base your test off of.

matthewp:
@pYr0x I think this is not needed anymore, right?

pYr0x:
i have to check if a multi main with npm package is working...

somethink like this
```

var promise = stealTools.build({
  main: ["myproject/main", "myproject/login"],
  config: __dirname+"/package.json!npm"
},{
  bundleSteal: true,
  minify: true,
  debug: false,
  quiet: false
});
```

matthewp:
Oh, I thought you had added a test like that. Ok, let me know.

pYr0x:
multi-main now works in minor 
still have some problems, but not with the normalization so close this

PR #372, opened 2016-02-14
Remove dependency on babel

### LINKED ISSUE #363: steal's babel 5 is trying to load .babelrc (with custom transpile option)
There is `.babelrc` with `Babel 6 options` in project folder. Custom `traspile` option is passed to steal-tools build. But it still tries to load `.babelrc.` and fails:

```
[BABEL]  Unknown option: .babelrc.presets 
```

matthewp:
This was added when I implemented the Trace extension. Isn't needed, Steal is able to load the browser-based babel and use that for transpiling just fine. Closes #363

### DIFF
--- a/lib/trace.js
+++ b/lib/trace.js
@@ -1,5 +1,3 @@
-var nodeRequire = require;
-
 
 var trace = function(System, BuildSystem, onFulfilled){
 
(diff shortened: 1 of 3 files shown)


whitecolor:
+1

Issue #142, opened 2014-12-20
Streaming API


matthewp:
It would be nice if we had a lower-level API that operated on Streams.  This would be particularly useful for the Watch mode, where a Promise doesn't make sense.  I propose dividing the API into something like this:

```js
stealTools.createBuildStream(system, options) -> Stream<Bundle>
```

```js
stealTools.build(system, options) -> Promise
```

```js
stealTools.continuousBuild(system, options) -> Stream<Bundle>
```

`createBuildStream` would be the low-level API that just returns a Stream containing Bundles.  `build` would work the same way it does today, except it would call `createBuildStream` and when bundles are received, write those to the file-system.  `continuousBuild` would do the same, but would just never resolve.

Same structure for pluginify.

whitecolor:
How is it going? Not ready yet?

matthewp:
Internally it's all streams, this is what `stealTools.build` is: https://github.com/stealjs/steal-tools/blob/master/lib/build/multi.js#L14

I think for your use case is you don't want the bundles to be written, right?  In that case the *easy* thing would be to simply add an option writeBundles: false or something and return the `buildStream`.

whitecolor:
Right, just to have only in memory result.

matthewp:
Yes, this would be very simple with the way the code works now.  I would add these lines: https://github.com/stealjs/steal-tools/blob/master/lib/build/multi.js#L13-L16

to another file, multi_stream.js maybe, and have it return the buildStream. and this could be `stealTools. createBuildStream` as this issue proposes.

whitecolor:
What about buding building fs output?

matthewp:
> What about buding building fs output?

That won't be done in this API, you'll get a stream of bundles and will be responsible for writing them out yourself.  We can expose a `createBundlesWriteStream` or something that takes a bundles stream and writes it out.  So you could create a bundles stream, manipulate it as you wish, and then write out the results.

whitecolor:
steam is ok.

matthewp:
Going to document what we need:

* **createBundleGraphStream** - Given

PR #374, opened 2016-02-16
Implement the streaming API

matthewp:
This is the intial streaming API, closing #142

The 4 streams are:

```js
stealTools.createGraphStream()
stealTools.createMultiBuildStream()
stealTools.createConcatStream()
stealTools.createWriteStream()
```

These need to be used in conjuction but allow you to swap in your own streams in the process to do other behaviors (or you could, say, omit the write stream to prevent writing to the file system).

Here's an example usage, this is basically what `stealTools.build` does:

```js

var createGraphStream = stealTools.createGraphStream;
var multiBuild = stealTools.createMultiBuildStream;
var concat = stealTools.createConcatStream;
var write = stealTools.createWriteStream;

var system = {
};

var stream = createGraphStream(system)
    .pipe(multiBuild())
    .pipe(concat())
    .pipe(write());
```

This PR includes docs and a guide that explains the usage better, see `doc/guides/streams.md`.


matthewp:
No plans to merge this immediately, want to leave it open for feedback for a while.

matthewp:
Thinking about moving these onto a `streams` object so that it becomes:

```js
var s = require("steal-tools").streams;

var stream = s.graph(config, options)
  .pipe(s.multiBuild())
  .pipe(s.concat())
  .pipe(s.write());
```

That's a bit nicer. When ES6 arrives it will be:

```js
const {
  graph,
  multiBuild,
  concat,
  write
} = require("steal-tools").streams;
```

matthewp:
Also, going to rename the multiBuild stream into `transpileAndBundle`.  I'd like to break these into separate streams but that might take a bit of worth. This at least names it correctly, when the stream is split into two it will be easier to deprecate.

PR #386, opened 2016-02-28
Bundle false

pYr0x:
this PR is a feature request. https://github.com/stealjs/steal-tools/issues/14
it is one of the oldest requests for steal-tools.

i made the possibility to ignore modules from the bundle process. 
you can now set jquery to a CDN for production and exclude jquery from node_modules.

Sorry for that jquery is inside the PR, but i dont know how to test with jquery CDN wihout having a local jquery version installed in the test node_modules-folder.

suggestion are welcome.

2 thing would be nice, maybe @justinbmeyer or @matthewp can help me with that.

1. if jquery is set to bundle=false in package.json, it would be nice if steal knows that and set a mapping to @empty if no mapping is provided in the production ENV
2. my problem when i run into this issue, was that steal-sass (transpiling) for building an app take very long. the problem is, that the sass.js module is about 2 MB. can we prevent modules from transpiling?


pYr0x:
@matthewp i am a little bit confused.. if found this PR https://github.com/stealjs/steal-tools/commit/64fc0cf76cd2e554759e03ad18650d031255980d
and it seems, that bundle: false was or is supported for a package.json config

so my work was vain?
that tests i wrote, should i use them to build a new PR that based on your bundle:false ans side bundle?

matthewp:
No, that was an old commit that later became sideBundle which is something different. Your PR is not in vain :)

pYr0x:
hm ok.. 
you can look over me PR and say if it is ok or i can do something better.
after i looked into your sideBundle PR my code for the bundle=false feature looks very dirty... :(

lib/graph/make_graph.js
```diff
@@ -90,8 +92,27 @@ module.exports = function(config, options){
 	});
 
 	return buildPromise.then(function(){
-		// When the build has loaded, update its config with any `buildConfig`
-		// options found.
+
+		var normalizedMetaPromises = _.map(configMetas, function(module){
```
matthewp: Why was this needed?  Was this for the test?

pYr0x: https://github.com/stealjs/steal-tools/pull/386/files#diff-e5bdcc156e0a2da2eb77e51a09c7b5ddR1687
in this test i have a package.json, that is loaded, i also add a config "meta"
but steal-tools load and normalize only the package.json, not the meta that pass though multibuild config.
if you remove my lines of code and run the test it("merge meta data with package.json", function(done){ jqueryt-module would not have bar=foo in the metadata

matthewp:
@pYr0x can you add your `ignore` option to the docs here: https://github.com/stealjs/steal-tools/blob/master/doc/types/build-options.md

pYr0x:
@matthewp updated some docs.

i ask in gitter, but can you explain how i can link (in markdown) to a documentation site -> to a h3 tag
like a html ancor tag #foobar

matthewp:
I think you can just do `[title](#foobar)`

pYr0x:
i do not understand...
i want a direct link to http://stealjs.com/docs/steal.html -> `<h3>Configure</h3>`

matthewp:
Open it in the devtools and find the id, if there is one, and just add a full link.

matthewp:
@pYr0x just took a look and that h3 doesn't have an id, so you can't link to it.

pYr0x:
whats about this http://stackoverflow.com/questions/6695439/how-do-you-create-link-to-a-named-anchor-in-multimarkdown/7015050#7015050
those your documentjs provide that?

matthewp:
I'm not sure how documentjs decides what gets an id and what does not, maybe only h1s and h2s get them.

pYr0x:
ok, so we can merge that.
i will look into documentjs or ask @daffl.
we can update the doc later

matthewp:
@pYr0x Can you push this branch to the steal-tools repo?  I have a change that removes the need for the meta normalization.  If you can push your branch to steal-tools I'll add my commit.

matthewp:
This is replaced by #389

pYr0x:
@matthewp i created new PR into steal-tools bundle-false repo. if you want you can merge that and add your commit into... let me know i need help...

matthewp:
Thanks, now it is in #390

PR #400, opened 2016-03-10
Bundle naming

pYr0x:
renaming bundles if npm package is provided.

the problem is, if using package.json, there is this `System.directories.lib` (usually `src`). this works great in development mode
```
<script 
	src="node_modules/steal/steal.js"
	env="window-development"
	main="mypkg/main"
></script>
```
-> file will load from `/src/main.js"

but in production mode it would fail with the old naming behaviour.
-> tried to load file from `/dist/bundles/mypkg/main.js`
but if we have this simple build:
**package.json**
```
..
"system": {
    "main": "app_a"
}
...
```

**build.js**
```
var s = require("steal-tools").streams;

var graphStream = s.graph({
	config: __dirname + "/package.json!npm",
});

var stream = graphStream
.pipe(s.transpile())
.pipe(s.minify())
.pipe(s.bundle())
.pipe(s.concat())
.pipe(s.write());
```
the file is located at `/dist/bundles/main.js`

---

with the new naming, a bundle like `mypkg@1.0.0#main`is now located at `/dist/bundles/mypkg/main.js`

add tests for nameing bundles with and without npm packages

### DIFF
--- a/lib/stream/bundle.js
+++ b/lib/stream/bundle.js
@@ -99,6 +99,11 @@ function bundle(data){
 	// Combine all bundles
 	var allBundles = splitMainBundles.concat(splitBundles);
 	allBundles = allBundles.concat.apply(allBundles, unbundledBundles);
+
+	// in each multibuild we will start from beginning.
+	// with that, it is possible now, that file collisions are present
+	// in different multibuilds
+	nameBundle.clearBundleCounter();
 	// Name each bundle so we know what to call the bundle.
 	allBundles.forEach(nameBundle);
 
--- a/test/bundle_name_test.js
+++ b/test/bundle_name_test.js
@@ -1,0 +1,184 @@
+var assert = require("assert"),
+	crypto = require('crypto'),
+	winston = require('winston'),
+	nameBundle = require("../lib/bundle/name");
+
+var dirName = "bundles/";
+
+describe("bundle name without npm package", function() {
+	beforeEach(function () {
+	});
+
+	it("works with just a normal modulename", function () {
+		var bundle = {
+			bundles: ["main"]
+		};
+		var bundleName = nameBundle.getName(bundle);
+		assert.equal(bundleName, dirName + 'main');
+	});
+
+	it("remove trailing parts", function () {
+		var bundleName = nameBundle.getName({bundles: ["main.js"]});
+		assert.equal(bundleName, dirName + 'main');
+
+		bundleName = nameBundle.getName({bundles: ["main/"]});
+		assert.equal(bundleName, dirName + 'main');
+
+		//bundleName = nameBundle.getName({bundles: ["foo@1.0.0#bar/"]});
+		//assert.equal(bundleName, dirName + 'bar');
+
+	});
+
+	it("works with multiple bundles", function(){
+		var bundle = {
+			bundles: ["main/bar", "bar/foo"]
+		};
+		var bundleName = nameBundle.getName(bundle);
+		assert.equal(bundleName, dirName + 'bar-foo');
+
+		bundle = {
+			bundles: ["main/bar", "bar/foo"],
+			buildType: "txt"
+		};
+		bundleName = nameBundle.getName(bundle);
+		assert.equal(bundleName, dirName + 'bar-foo.txt!');
+	});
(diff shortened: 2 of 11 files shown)


matthewp:
Thanks, don't worry about the build failing for now. Let's talk about this at standup.

pYr0x:
all right, hopefully i did a good job ;)

test/bundle_name_test.js
```diff
@@ -0,0 +1,184 @@
...
+
+	it("remove trailing parts", function () {
+		var bundle = {
+			bundles: ["foo@1.0.0#main/bar/", "foo@1.0.0#foo.js"]
+		};
+		var bundleName = nameBundle.getName(bundle);
+		assert.equal(bundleName, dirName + 'foo/bar-foo');
+	});
+
+	it("works with a plugin markup", function(){
```
matthewp: This is the one I want to make sure is still working like it previously was.  I think that it is.

matthewp:
@pYr0x My opinion on the combined bundles (those with 2 or more bundles) is that we can remove the package name and have it be flattened.

pYr0x:
Naming Rules
===

_bundle-folder = "/bundles"_
_[[HASH]] = a unique hash string build from the concatenated bundles string_
_[[COUNTER]] = if the same file exists (in one build), increase the counter_

## without NPM
### main bundle only

| module | bundled path | note |
| --- |--- | --- |
|  mymodule | `/bundles/mymodule.js`
|  mymodule.js | `/bundles/mymodule.js`
|  mymodule | `/bundles/mymodule.js`
| deep/folder/structure/with/very/very-very-very-long-name | `/bundles/deep/folder/structure/with/very/very-very-very-long-name.js` | main bundles arent cut
| index.stache!done-autorender | `/bundles/index.js`
| mymodule!my-plugin | `/bundles/mymodule.js`

### with bundles
| module | bundled path | note |
| --- |--- | --- |
`["main/bar", "bar/foo"]` | `/bundles/bar-foo.js` |
`["main/bar", "bar/foo"]` | `/bundles/bar-foo.js` |
| `["main/my-very-long-name", "bar/eman-gnol-yrev-ym"]` | `/bundles/my-very-long-nam-[[HASH]].js`| [[HASH]] = a unique hash string, concat bundle[0] + bundle[1]  cut to 25 chars |
| `["main/my-very-long-name", "bar/eman-gnol-yrev-ym"]` | `/bundles/my-very-long-nam-[[HASH]]-[[COUNTER]].js`| [[HASH]] = a unique hash string; concat bundle[0] + bundle[1]  cut to 25 chars; if the same file exists, put a counter var at the end |
| `["deep/folder/structure/with/very/very-very-very-long-name", "deep/space/nine/is/a/great-series/from-the-90s"]` | `/bundles/very-very-very-l-f802f7c2-[[HASH]].js` | [[HASH]] = a unique hash string; concat bundle[0] + bundle[1]  cut to 25 chars |
`["main/bar0815.com.js", "bar/foo- bar"]` | `/bundles/bar0815-com-foo-bar.js` | for bundles, we only return filenames with chars, numbers, `-` and `_`

lib/stream/bundle.js
```diff
@@ -99,6 +99,11 @@ function bundle(data){
 	// Combine all bundles
 	var allBundles = splitMainBundles.concat(splitBundles);
 	allBundles = allBundles.concat.apply(allBundles, unbundledBundles);
+
+	// in each multibuild we will start from beginning.
+	// with that, it is possible now, that file collisions are present
+	// in different multibuilds
+	nameBundle.clearBundleCounter();
```
pYr0x: @matthewp i added this function, because if not, the tests will fail.
the problem is that the tests run mutliple build in one task. currently the `usedBundleNames` is leaking.

there are 2 problems with `usedBundleNames`

* if its not leaking, we cant cover file collisions with multiple builds in one task
* if it is leaking and we remove the `nameBundle.clearBundleCounter();` the tests will fail (https://github.com/stealjs/steal-tools/blob/master/test/multibuild_test.js#L2004) because the test to not check the `bundleCounter`

matthewp:
Closed in favor of #404

PR #402, opened 2016-03-13
Bundle naming none leaking

pYr0x:
this is a none leaking version of name.js

with this version, we can not cover file collisions over multiple builds

### DIFF
--- a/lib/build/transform.js
+++ b/lib/build/transform.js
@@ -120,7 +120,7 @@ var transformImport = function(config, transformOptions){
 				nodes: nodesInBundle,
 				buildType: nodesInBundle[0].load.metadata.buildType || "js"
 			};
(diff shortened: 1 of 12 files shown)


matthewp:
Great work here, going to merge, would you mind taking the table you made [here](https://github.com/canjs/can-zone/tree/v0.2.14#ignore) and putting it somewhere... maybe `lib/bundle/name.md` just so we can reference it.

matthewp:
Actually @pYr0x can you resubmit this PR against the `minor` branch instead of master? This will go into the next minor release.  Want to keep master for bug fixes that can be part of 0.15.x

pYr0x:
close https://github.com/stealjs/steal-tools/pull/404

PR #404, opened 2016-03-16
Bundle naming none leaking

pYr0x:
@matthewp please merge against minor


matthewp:
Great, thanks, will you put that table somewhere?

PR #406, opened 2016-03-17
adding naming rules as a comment

pYr0x:
so we can use it for a review if we need 

@matthewp is it ok?


matthewp:
Not so readable :(  Why not as a `.md` file? Will show up on github nicely.

pYr0x:
name.md in the bundle folder ?

matthewp:
Yeah.

PR #410, opened 2016-03-17
NPM multi-main build

pYr0x:
this PR is not ready.

we have to work on:

- [x] add test with single main **pass**

- [x] add test with a main in a single array **fail**

- [ ] add test with multi-main

- [ ] resolve problem

PR is against minor


matthewp:
Closed in favor of 413

PR #417, opened 2016-03-18
NPM multi main

pYr0x:
this PR is not ready.

we have to work on:

- [x] add test with single main **pass**

- [x] add test with a main in a single array **pass**

- [x] add test with multi-main **pass**

- [x] resolve problem

- [ ] multi-main with `bundleSteal=true`, **what should be the result?** all dependencies are bundled no matter if they can be loaded progressive? currently, they same algorithm is used as `bundleSteal=false` but the progessive loaded files cant be loaded, because there is not information in the bundled file where to load them. **failed**
cc @justinbmeyer @matthewp 

PR is against minor

### DIFF
--- a/lib/graph/make_graph_with_bundles.js
+++ b/lib/graph/make_graph_with_bundles.js
@@ -135,7 +135,7 @@ var makeBundleGraph = module.exports = function(config, options){
 				copy.main = nextBundle;
 				return dependencyGraph(copy, options).then(function(data){
 					mergeIntoMasterAndAddBundle(masterGraph, data.graph,
-												nextBundle, data.loader);
+												data.steal.System.main, data.loader);
 					mergeConfigForNextBundle(options.system, data.loader);
 					return getNextGraphAndMerge();
 				});
--- a/lib/graph/make_graph_with_bundles.js
+++ b/lib/graph/make_graph_with_bundles.js
@@ -111,16 +111,25 @@ var makeBundleGraph = module.exports = function(config, options){
 			var nextBundle = bundleNames.shift();
 
 			if(!nextBundle) {
-				// If there are no more bundles, return data
-				return {
-					graph: masterGraph,
-					steal: data.steal,
-					loader: data.loader,
-					buildLoader: data.buildLoader,
-					mains: data.mains,
-					config: config,
-					options: options
-				};
+				var mains = Array.isArray(config.main) ? config.main.slice(0) :
+					[data.loader.main];
+
+				return Promise.all(
+					mains.map(function(main){
+						return data.loader.normalize(main);
+					})
+				).then(function(mains){
+					// If there are no more bundles, return data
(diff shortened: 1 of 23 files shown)


lib/graph/make_graph_with_bundles.js
```diff
@@ -135,7 +135,7 @@ var makeBundleGraph = module.exports = function(config, options){
 				copy.main = nextBundle;
 				return dependencyGraph(copy, options).then(function(data){
 					mergeIntoMasterAndAddBundle(masterGraph, data.graph,
-												nextBundle, data.loader);
+												data.steal.System.main, data.loader);
```
matthewp: Can you explain why this is needed. Is it because nextBundle is unnormalized?

matthewp:
I don't think bundleSteal is a great idea with multi-main. It would wind up adding steal to each of your mains which puts a lot of extra weight into each of them.

There is already a test for bundleSteal with multi main here: https://github.com/stealjs/steal-tools/blob/master/test/multibuild_test.js#L1293  This is not using the npm plugin or directories.lib, but it does pass.

lib/graph/make_graph_with_bundles.js
pYr0x: yes... nextBundle is not normalized

pYr0x:
yes i know it is not best prectis, but we should cover that sice the test without NPM is working (https://github.com/stealjs/steal-tools/blob/master/test/multibuild_test.js#L1293) and with NPM it is not (https://github.com/stealjs/steal-tools/pull/417/files#diff-e5bdcc156e0a2da2eb77e51a09c7b5ddR1821)

matthewp:
Yeah, definitely should cover it with a test.  By the way, I don't think the npm plugin is the reason for this so much as is the use of `directories.lib`. Using the npm plugin should be more or less like without it (like the current passing test), adding directories.lib is what causes all of the issues we frequently encounter.

pYr0x:
@matthewp this test (https://github.com/stealjs/steal-tools/pull/417/files#diff-e5bdcc156e0a2da2eb77e51a09c7b5ddR1821) fails because
* `baseUrl` is overwritten, with this PR not anymore https://github.com/stealjs/steal/pull/633#issuecomment-199517208
* `System.bundles` is empty in bundled file with NPM, but not with this test (https://github.com/stealjs/steal-tools/blob/master/test/multibuild_test.js#L1293)... strange

matthewp:
Hey @pYr0x, I'm working on this stuff now. I can see that the problem is that the bundles are not added to the `System.bundles` configuration. Going to add some unit tests to `lib/bundle` to test the creation of the bundles config.

pYr0x:
@matthewp great, later today we can talk about that.

matthewp:
@pYr0x if you want to try this out and let me know how things go in your own app.

Issue #419, opened 2016-03-20
bundleSteal env is by default: development


pYr0x:
if you make a build and set the option `bundleSteal=true` and load your app like `<script src="dist/bundles/testapp/main.js" main="testapp/main"></script>`, the env by default is "window-development" which is wrong.

matthewp:
So, you might be thinking of adding something like:

```js
steal.env = "window-production";
```

While that probably works and is probably ok... I don't love the idea because it prevents us from using the bundled file somewhere other than in a web-browser (like running on a server).

Maybe instead we could add an option like: `steal.stealBundled = true` to the bundled file and then the same way we check for `steal.production` we could check for this property.  Here: https://github.com/stealjs/steal/blob/master/src/config.js#L232

pYr0x:
@matthewp i like that way with `steal.stealBundled = true` ...

do we need that also if steal is bundled ?
https://github.com/stealjs/steal/blob/master/src/config.js#L230
i think we can leave setting `stealPath()` and the whole `paths` configs. or do i miss something?

matthewp:
I think that might be ok. Try it out and let me know what happens.

pYr0x:
https://github.com/stealjs/steal-tools/pull/425

PR #423, opened 2016-03-21
let steal know if it is bundled

pYr0x:
if steal is bundled we should cover that and dont overwrite `steal.baseUrl` which is set in the bundled file
related to https://github.com/stealjs/steal-tools/issues/419


pYr0x:
@matthewp tests pass.

if added steal in package.json to my private fork `"steal": "git://github.com/pYr0x/steal.git#stealBundled",` can you please set this to the correct steal build

hopefully my branch chaos is not entangle you

matthewp:
Looks good, thanks, will do a release shortly.

PR #427, opened 2016-03-30
Add tests to cover all multi-main situations

matthewp:
Thanks to @pYr0x for doing all of these. This replaces #417

Now all of the multi-main situations are covered.


matthewp:
This also merges master into minor.

PR #432, opened 2016-04-05
Refactor CLI.

m-mujica:
screen shot 2016-04-05 at 16 20 42

screen shot 2016-04-05 at 16 21 09

screen shot 2016-04-05 at 16 21 29

download

### DIFF
--- a/bin/steal
+++ b/bin/steal
@@ -20,7 +20,7 @@ var argv = require("yargs")
 
 // run `steal build` by default if no command was provided
 if (!argv._[0]) {
-	var spawn = require("child_process").spawn;
+	var spawn = require("cross-spawn-async");
 	var binary = path.join(__dirname, "steal");
 	var rawArgs = ["build"].concat(process.argv.slice(2));
 
--- a/bin/steal
+++ b/bin/steal
@@ -1,179 +1,31 @@
 #!/usr/bin/env node
-var winston = require('winston');
-var path = require("path");
-var _includes = require("lodash/includes");
-var commands = ["build", "transform", "export", "live-reload"];
 
-var yargs = require("yargs")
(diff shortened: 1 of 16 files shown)


bin/steal
```diff
@@ -1,179 +1,31 @@
...
+var cmdLiveReload = require("../lib/cli/cmd_live_reload");
+
+var argv = require("yargs")
+	.command(cmdBuild)
+	.command(cmdTransform)
+	.command(cmdLiveReload)
+	.command(cmdExport)
+	.version(function() {
+		return require("../package.json").version;
+	})
```
matthewp: Lazily loading the package.json, I love it!

bin/steal
```diff
@@ -1,179 +1,31 @@
...
-		doLiveReload(options);
-		break;
-}
+	.argv;
 
-function build() {
-	var promise = stealTools.build(system, options);
+// run `steal build` by default if no command was provided
+if (!argv._[0]) {
+	var spawn = require("child_process").spawn;
```
matthewp: We need to use cross-spawn-async for Windows.  I can update this.

matthewp:
I'm going to get this to pass in Windows, then will merge.

bin/steal
m-mujica: I originally used cross-spawn-async here (and also replace the one in the test_cli file) but it brakes the build in 0.10.x because cross-spawn-async uses or depends on something that uses `Symbol`. Hopefully there is a way to workaround this limitation.

matthewp: Ah, shucks, ok, in that case I think we can do this: https://github.com/stealjs/steal-tools/blob/master/test/test_cli.js#L24  

I'll update it.

PR #447, opened 2016-04-16
Make sure apps using steal-conditional build properly.

m-mujica:
it should be good to go now.....

high_five_stephen_colbert


matthewp:
Looks good, I released steal-conditional 0.1.3 with your fix.

PR #457, opened 2016-04-28
Fixes #456. Throw error on build promise rejection

### LINKED ISSUE #456: Any error the stops steal-tools from building should also set an exit code greater than 0
```bash
work:bithub-admin mark$ ./node_modules/steal-tools/bin/steal build --main=bithub-admin/bithub-social
OPENING: bithub-admin/bithub-social
Potentially unhandled rejection [6] Error: Error loading "style/style.less!$less" at file:/Users/mark/Projects/BithubApp/bithub-admin/style/style.less
Error loading "style/style.less!$less" from "bithub-admin@0.1.0#bithub-social" at file:/Users/mark/Projects/BithubApp/bithub-admin/bithub-social.js
variable @grid-columns is undefined
work:bithub-admin mark$ echo $?
0
```

From the above example, you can see the build was not successful due to an error loading `style/style.less`. But when I echo the exit code of the command it returns a `0`.

mjstahl:


matthewp:
Great, thanks, will release tomorrow if that's ok. Just ping me.

PR #461, opened 2016-05-11
Add support for load.metadata.includedDeps

matthewp:
This adds support in live-reload for a new property:

```js
load.metadata.includedDeps = [
  "path/to/module"
];
```

This can be set by things like plugins, and this data is used to keep track of dependencies that are not part of the main dependency graph. live-reload will still track these files and emit for the parent module when the `includedDeps` changes.

In the client we now need to support receiving an array of module names. This is because includedDeps are not part of the real dependency graph, and therefore can trigger reloads in multiple parents.


matthewp:
Not ready to merge this yet, need to update live-reload so it can accept an array of moduleNames to reload.

Will release both steal/steal-tools as pre-releases (under tags) so the entire flow, with less can be tested.

Issue #443, opened 2016-04-11
Global shim support exporting global.  


justinbmeyer:
Often you want a module to export a global.  Currently, I hack that in like this:

```
if(typeof window !== "undefined" && !require.resolve) {
	window.fixture = fixture;
}
```

We should make `exports` be able to do the reverse.  `{"can-fixture": "fixture"}`

justinbmeyer:
The way to do this is to make `lib/bundle/shim.js`'s `define` after the export is saved here:

```
modules[moduleName] = module && module.exports ? module.exports : result;
```

it should check if `moduleName` is in `exports` and if it is, add the right thing to the global. We probably need a `set` similar to `shim.js`'s `get` that could set a nested property like:

```
exports: {"can-fixture", "can.fixture"}
```

I wonder if we could even supply multiple.  That would be cool for it to try to put it on `window.can`, failing that `window`.

matthewp:
This is available now: https://github.com/stealjs/steal-tools/pull/463

PR #463, opened 2016-05-17
Use exports to specify global exported values

matthewp:
This updates transform's `exports` property so that it can be used to specify values that should be exported.

For example if you have a module **foo** that looks like:

```js
module.exports = "foo bar";
```

And you set exports:

```js
exports: {
  "foo": "foo.bar"
}
```

Then it will set it so that `window.foo.bar === "foo bar"`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ test/pluginifier_builder_helpers/browserify-out.js
 test/6to5/dist
 test/bundle_assets/dist
 test/live_reload/out.js
+test/transform_export/out.js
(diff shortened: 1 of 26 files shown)


pYr0x:
what is that good for? what is the usecase of that feature?

matthewp:
@pYr0x canjs will use this feature for 3.0.  An example would be can/component. For the global build you want that module to be placed on `can.Component` on the window.  Without this feature you have to do this manually in your code.

matthewp:
Fixes #443

PR #471, opened 2016-06-13
Upgrade for Node 6 compatibility

### LINKED ISSUE #455: Warnings with Node 6
When running a steal-tools build the following warnings (among others) are shown:

```
(node:14137) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.
```

matthewp:
Fixes #455

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "grunt-cli": "^0.1.13",
     "grunt-contrib-concat": "~0.3.0",
     "grunt-contrib-copy": "~0.5.0",
-    "grunt-contrib-jshint": "^0.10.0",
+    "grunt-contrib-jshint": "^1.0.0",
     "grunt-contrib-uglify": "~0.2.2",
     "grunt-contrib-watch": "~0.3.0",
     "grunt-release": "^0.7.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "grunt-cli": "^0.1.13",
     "grunt-contrib-concat": "~0.3.0",
     "grunt-contrib-copy": "~0.5.0",
-    "grunt-contrib-jshint": "^0.10.0",
+    "grunt-contrib-jshint": "^1.0.0",
```
daffl: Can we get rid of Grunt the same way we did for all the other modules?

matthewp: I don't think so, we have a grunt task we use to test our grunt tasks.

matthewp: We could maybe move the grunt tasks to a different repo/npm though.

daffl: Keeping it to test the task is fine. I was just thinking about JSHint, Testee and the release task.

matthewp: Yeah, we can do this at any time.

PR #475, opened 2016-06-16
work with steal promise based startup

pYr0x:


pYr0x:
close.. should be merged against major branch!

PR #476, opened 2016-06-16
work with steal promise based startup

pYr0x:
this PR adds the new steal

👍 👍 👍 👍 👍 

@matthewp pls short review, thanks

* change some test in `dependencygraph_test.js`, they should configured like the multibuild test
* we removed `root` and `startId` in steal config, so we renamed there in those tests

### before merge, change the steal repo on package.json and bower

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "steal-tools",
   "version": "1.0.0-beta.0",
   "devDependencies": {
-    "steal": "^1.0.0-beta.0",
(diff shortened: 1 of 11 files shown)


matthewp:
Not sure if you left the `.only` in on purpose or not?

pYr0x:
right... thanks

matthewp:
@pYr0x it's failing in appveyor because of something that is fixed in major.  If you update this line in your package.json: https://github.com/stealjs/steal-tools/blob/major/package.json#L116  It should pass.

Issue #484, opened 2016-06-26
[node v6.2.2] [npm  v3.9.5] sh: install-engine-dependencies: command not found


leoj3n:
Project source: https://github.com/stealjs/steal/wiki/Getting-Started

`npm install`:
~~~
npm ERR! node v6.2.2
npm ERR! npm  v3.9.5
~~~

`package.json`: (notice `"steal": "stealjs/steal#major"`)
~~~
{
  "name": "myhub",
  "version": "1.0.0",
  "description": "",
  "main": "myhub.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "bootstrap": "^3.3.6",
    "jquery": "^3.0.0",
    "justifiedGallery": "^3.6.2",
    "steal": "stealjs/steal#major"
  },
  "devDependencies": {
    "steal-qunit": "^0.1.1",
    "steal-tools": "stealjs/steal-tools#major"
  },
  "system": {
     "map": {
         "justifiedGallery": "justifiedGallery/src/js/justifiedGallery"
     },
     "meta": {
         "justifiedGallery/src/js/justifiedGallery": {
             "format": "global",
             "deps": ["jquery","justifiedGallery/src/less/justifiedGallery.less"]
         }
     }
  }
}
~~~

`npm install`: (notice `sh: install-engine-dependencies: command not found`)
~~~
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue

> fsevents@1.0.12 install /Users/leoj/workspace/stealjs/myhub/node_modules/fsevents
> node-pre-gyp install --fallback-to-build

gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
gyp WARN download NVM_NODEJS_ORG_MIRROR is deprecated and will be removed in node-gyp v4, please use NODEJS_ORG_MIRROR
  SOLINK_MODULE(target) Release/.node
  CXX(target) Release/obj.target/fse/fsevents.o
  SOLINK_MODULE(target) Release/fse.node
[...]

PR #485, opened 2016-06-26
make engine-dependencies work

pYr0x:
`postinstall: install-engine-dependencies` does not work if the module is not installed globally.
see also
https://github.com/rxaviers/cldr-data-npm/issues/33#issuecomment-227923754

@matthewp with `prepublish` now `zombi` is not installed on the users local machine. 
we should really make installing steal-tools more lighter. nobody need `bower` if using steal-tools for building apps. https://github.com/stealjs/steal-tools/blob/major/package.json#L37

should close https://github.com/stealjs/steal-tools/issues/484

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,8 @@
     }
   ],
   "scripts": {
-    "postinstall": "install-engine-dependencies",
+    "postinstall": "npm run install-engine-dependencies",
+    "install-engine-dependencies": "install-engine-dependencies",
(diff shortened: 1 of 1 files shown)


matthewp:
Thanks! Will update this Monday.

matthewp:
I think we just need to move engine-dependencies as a regular dep.

matthewp:
Closing in favor of https://github.com/stealjs/steal-tools/pull/488

PR #486, opened 2016-06-26
update to newest clean-css will minify with correct order

### LINKED ISSUE #467: Invalid order of styles after build
Here is small example
https://github.com/Krakabek/StealStylesBug
All styles are in one file: styles.less
After building (node build.js) you can find index.css in dist folder
You can see, that body,input{font-family:"Helvetica Neue"... goes before reset block, containing font:inherit
Since order is critical in css files, produced app will have font:inherit in body

pYr0x:
close #467

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "chokidar": "^1.0.1",
-    "clean-css": "3.1.8",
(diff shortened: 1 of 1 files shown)


matthewp:
Thanks!

PR #488, opened 2016-06-27
Move engine-dependencies to a regular dep

matthewp:
This will cause it to run whenever a `npm install` completes. If the user is using steal-tools (not developing steal-tools) then nothing will happen. If they are developing steal-tools, engine-dependencies will install the correct version of zombie for the Node version.


pYr0x:
> If the user is using steal-tools (not developing steal-tools) then nothing will happen

where is this set?

matthewp:
> where is this set?

They're marked as devDependencies: https://github.com/stealjs/steal-tools/blob/master/package.json#L84

And engine-dependencies knows not to install them: https://github.com/bitovi/engine-dependencies/blob/master/engine-dependencies.js#L17-L22

matthewp:
Released as 1.0.0-beta.1

PR #494, opened 2016-07-18
Add websocket ssl (wss) support to live-reload

mjstahl:
JSHint

Add tests

Refactor tests to not throw assertion exception


matthewp:
Had to do some rebasing so this is moved to #501

PR #506, opened 2016-07-30
configless steal multi-main bundle

pYr0x:
**a multi-main build with steal bundled**
- should not needed to specify a `main` in script tag
- should work without config, just paste the script tag like `<script src="./dist/bundles/multi-main-bundled/app_a.js"></script>`

_added `steal.main` to the `add_steal.js`_


matthewp:
Nice easy win 👍

Issue #507, opened 2016-08-01
Specifying System.bundle with wildcards does not work with package.json during the build process...


georgesunil81:
I have some pages and modals in my CanJS application that dynamically get loaded using System.import

**I have the System.bundle specified as below in my package.json -**

```
"system": {
"bundle": [
      "my-app/my-pages/**/*",
      "my-app/my-modals/**/*"
]``
```

**I then run the build process (node build.js), it gives me these errors -**

 Error: Error loading `"my-app/my-pages/**/*"`. ENOENT: no such file or directory.

**Below is my build.js -**

```
var stealTools = require("steal-tools");

stealTools.build({
  config: __dirname + "/package.json!npm",
  bundlesPath: "dist/bundles/my-app"
}, {
  bundleAssets: true
});
```

It currently builds fine if I list each bundle individually within System.bundle, but I run the risk of forgetting to add to it if I add some more pages or modals to my application! We'd like the ability to specify asterisks within the System.bundle so that it reads those in dynamically during the build process.

frank-dspeed:
@matthewp  this can be closed

PR #512, opened 2016-08-17
bundle parameter now handels NPM modules and globs

pYr0x:
## WIP

first commit for https://github.com/stealjs/steal-tools/issues/507

### DIFF
--- a/lib/loader/find_bundle.js
+++ b/lib/loader/find_bundle.js
@@ -26,11 +26,12 @@ function findBundles(loader) {
 	var patternArray = [];
 	var bundleArray = [];
 
-	// add direcotries.lib to globs current working directory
+	// add direcotries.lib to the current working directory
 	if(loader.npmContext && loader.directories && loader.directories.lib) {
 		globOpts.cwd = path.join(globOpts.cwd, loader.directories.lib);
 	}
-
+	
+	// create a patternArray
 	if(Array.isArray(loader.bundle)) {
 		patternArray = loader.bundle;
 	}else if(typeof loader.bundle === "string") {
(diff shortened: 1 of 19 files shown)


lib/loader/find_bundle.js
```diff
@@ -9,30 +14,126 @@ module.exports = findBundles;
...
-function minusJS(name) {
-	var idx = name.indexOf(".js");
-	return idx === -1 ? name : name.substr(0, idx);
+function endsWith(str, suffix) {
+	return str.indexOf(suffix, str.length - suffix.length) !== -1;
 }
+function startsWith(starts, path) {
+	return path.substr(0,starts.length) === starts;
+}
+function isNpm(moduleName){
```
matthewp: you can require npmUtils and get all of this stuff: `var npmUtils = require("steal/ext/npm-utils");`

pYr0x: ok. thanks.
i have to ivestigate the problem, why the tests are not passing.

matthewp: I think major might be failing in general.

pYr0x: why?

pYr0x:
@matthewp i think i found the problem
https://github.com/stealjs/steal-css/pull/19
can you make a release for steal-css and merge it into steal?

matthewp:
Yep, doing it now, will let you know and you can update this PR.

matthewp:
https://github.com/stealjs/steal/pull/829

pYr0x:
you also add the css.js into steal and make a new rc release?

matthewp:
in the process :)

matthewp:
@pYr0x rc-1 has this. Thanks!

pYr0x:
thanks @matthewp 
the tests are passing. but one is failing
```
1) multi build Source Maps works with configDependencies:
     Error: ENOENT: no such file or directory, open '/home/travis/build/stealjs/steal-tools/test/npm-config-dep/dist/bundles/main.js.map'
```

i think this is releated to https://github.com/stealjs/steal/issues/803 ?

should we merge that PR and open a new PR for stripping the `sourceMapping` ?

matthewp:
I think it would be more maintainable of we just removed the glob pattern and ran it through loader.normalize. That would handle the appName and ~/ conventions and all of that other stuff. I'm afraid this code will break accidentally if we change the npm plugin in the future.

I'm thinking it should be like:

1. Remove `/**/*` (or whatever it is).
2. `loader.normalize(name, loader.main)`
3. Add `/**/*` back on.
4. Do directories.lib stuff (sad that this is needed)
5. Then run through the glob finder.

This would make it so that any normalize hook stuff is covered, like the backslash extension.

pYr0x:
what to you mean by 
> Remove /**/*

we use not only the glob, we also use the [minimatch ](https://www.npmjs.com/package/minimatch) pattern, see the tests.
https://github.com/stealjs/steal-tools/blob/8b128f2f7accd7a4b24741b47882dbd02a41d33d/test/bundle_test.js#L14-L82

matthewp:
At this point we already have the dependency graph, it's in `data.graph`. Each key in the graph is the name of a module in the graph.  What I think we should do instead is try and match the bundles against items in that graph.  This would avoid going to the filesystem at all. 

What I really want to avoid is having any logic that checks for `~/` or the package name or any of that sort of stuff. steal-tools shouldn't have to know about `~/` at all, only the npm plugin should need to know about that.

That's why I was thinking that if you could turn a value like `app/components/**/*` and normalize that which becomes `app@1.0.0#components/**/*` you could then try to match each entry in the dependency graph which would find `app@1.0.0#components/tabs/tabs` for example. Make sense?

I know that we were normalizing it in that other function before, but I think that was a bad way to do it.

pYr0x:
Thanks for your answer.
You say that we have a dependency graph at this point, but a bundle entry point needn't be in the dependency graph. 
http://forums.donejs.com/t/whats-the-different-between-sidebundle-and-bundle/206

matthewp:
The dependency graph is everything that Steal loads. Nothing outside of the dependency graph gets written into a bundle (or a sideBundle) so we don't need to look at the filesystem, everything has already been fetched as this point.

pYr0x:
```
if(window.location.hash === "#login") {
      System.import("site/login").then(function(){
        $("#main").login();
      });
    } else if(window.location.hash === "#signup" ) {
      System.import("site/signup").then(function(){
        $("#main").signup();
      });
    } else {
      System.import("site/homepage").then(function(){
        $("#main").homepage();
      });
    }
```
E.g. Site/login will not be in dependeny graph, right?

matthewp:
Ugh, you're right, I was thinking this was only used after the full dependency graph was loaded. Arg, that not great.

In that case, I would rather just not accept the package name or `~/` to be included when using the `**/*` form. Let's throw an error if you try and use those. Either put real module names (like `app/components/foo` or use globs that go through the file system `components/**/*`. 

In this case I think we can remove the npm-specific like except perhaps the directories.lib part.

matthewp:
@pYr0x Is this something you're still interested in working on? No pressure, of course, just wondering if it's more work than its worth...

pYr0x:
i would like to finish it. but if have no idea when to do it ;)

matthewp:
No problem. 🤠

matthewp:
I'm going to close the PR for now. Feel free to reopen when it's ready.

PR #517, opened 2016-08-28
Strip sourcemap on sources

pYr0x:
strip source maps now while concatenating files.
also add to the sourcemap regex the `multline` parameter `/m`

this is related to: https://github.com/stealjs/steal/issues/803

with this fix, we now strip all sourcemaps on all files that will be concatenated.

### DIFF
--- a/lib/buildTypes/minifyJS.js
+++ b/lib/buildTypes/minifyJS.js
@@ -1,5 +1,4 @@
 var UglifyJS = require("uglify-js");
-var removeSourceMapUrl = require("../remove_source_map_url");
 
 module.exports = function(source, options){
 	var opts = (options != null) ? options.uglifyOptions : {};
@@ -18,7 +17,5 @@ module.exports = function(source, options){
 		}
 	}
 
-	var result = UglifyJS.minify(code, opts);
-	result.code = removeSourceMapUrl(result.code);
-	return result;
+	return UglifyJS.minify(code, opts);
 };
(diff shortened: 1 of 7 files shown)


lib/buildTypes/minifyJS.js
```diff
@@ -1,5 +1,4 @@
 var UglifyJS = require("uglify-js");
-var removeSourceMapUrl = require("../remove_source_map_url");
```
matthewp: I get why you added it to concat, but why did you remove it at this step?

pYr0x: as i saw, all build process goes through the steam API?
and as i saw, the concatenation is always after transpiling? 

https://github.com/stealjs/steal-tools/blob/2ff7f77322e70cbf7b33484387353ddbcaf088ae/lib/build/multi.js#L89
https://github.com/stealjs/steal-tools/blob/2ff7f77322e70cbf7b33484387353ddbcaf088ae/lib/build/multi.js#L94

do i missed something?

https://github.com/stealjs/steal-tools/blob/792ecdba1b4163518bd3ab3012bc748b4141f350/lib/build/transform.js#L110
https://github.com/stealjs/steal-tools/blob/792ecdba1b4163518bd3ab3012bc748b4141f350/lib/build/transform.js#L140

AND

matthewp: Oh I see now, because you added the multiline flag to the source map regex it will remove *all* source maps from the concatenated string, is that right?

pYr0x: yes.

i think that uglify will also strip the sourceMap https://github.com/mishoo/UglifyJS2/blob/41a9329409ca0f1273b6ccdcc9770713f44ec55a/test/mocha/input-sourcemaps.js

PR #535, opened 2016-10-26
Adds test showing ES6 circular dependencies working with Babel

### LINKED ISSUE #382: circular dependencies not resolving correctly in production build
foo.js
```javascript
import Map from 'can/map/';
import 'can/map/define/'; 
import bar from 'myproject/src/bar/';

export const foo = Map.extend({
  define: {
    propA: {
      get() { return bar.attr('propA'); }
    },
    propB: {
     get() { return 'Hello World!'; }
  }
});
```
bar.js
```javascript
import Map from 'can/map/';
import 'can/map/define/'; 
import foo from 'myproject/src/foo/';

export const bar = Map.extend({
  define: {
    propA: {
      get() { return 'Hello World!';  }
    },
    propB: {
     get() { return foo.attr('propB'); }
  }
});
```
this works in development. However in production,
the define for foo calls
```javascript
var _bar = _interopRequireDefault(_myprojectBar);
```
If the define for foo is called before the define for bar, then `_bar === { 'default': {} }`

After bar is define, it seems that _bar in the foo define is not updated so that when foo.attr('propA') is called, it returns an undefined TypeError

matthewp:
Closes #382

### DIFF
--- a/test/circular/main.js
+++ b/test/circular/main.js
@@ -1,0 +1,6 @@
+import { getBar } from './foo';
+import { getFoo } from './bar';
+
+// true in dev, false in production
+window.circularWorks = (getBar() === 'bar') &&
+	(getFoo() === 'foo');
--- a/test/circular/bar.js
+++ b/test/circular/bar.js
@@ -1,0 +1,7 @@
+import { getBar, name as fooName } from './foo';
+
+export let name = 'bar';
+
+export function getFoo() {
+	return fooName;
+}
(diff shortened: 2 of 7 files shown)


test/circular/main.js
```diff
@@ -0,0 +1,6 @@
+import { getBar } from './foo';
+import { getFoo } from './bar';
+
+// true in dev, false in production
```
pYr0x: What do you mean with that comment?
In the multibuild_test you check against true in prod.html

matthewp: Comment was old, when this was breaking it was true in dev and false in prod. It's working now, just forgot to remove the comment.

PR #564, opened 2016-11-28
Remove dev code from configuration graph

### LINKED ISSUE #562: Remove development code of configDependency module
`//!steal-remove-{start,end}` doesn't seem to work on config dependencies, debugging a little bit it seems the reason is the configDependency is not part of the [dependency graph passed to `stream/transpile`](https://github.com/stealjs/steal-tools/blob/f93a359739a213d51bd3cfdd8d183f8b295dd34a/lib/stream/transpile.js#L38), is this something we can fix/workaround?

m-mujica:
Closes #562


matthewp:
👍

Issue #579, opened 2017-01-11
socket.io-client/dist/socket.io works in the client, but not during a build.


justinbmeyer:
Related to: https://github.com/canjs/canjs/pull/2934

If an application's main looks like: `require("socket.io-client/dist/socket.io")`, it will be able to load in the client, but fail during a build like:

```js
var stealTools = require("steal-tools");

stealTools.build({
	main: "can/test/test",
	config: __dirname + "/../package.json!npm"
},{
	minify: false,
	debug: true,
	bundleAssets: true
});
```

It fails with:

```
Potentially unhandled rejection [5] TypeError: Error loading "socket.io-client@1.7.2#dist/socket.io" at file:/Users/justin/dev/canjs/node_modules/socket.io-client/dist/socket.io.js
Error loading "socket.io-client@1.7.2#dist/socket.io" from "can@3.3.1#test/test" at file:/Users/justin/dev/canjs/test/test.js
Multiple defines for anonymous module
    at /Users/justin/dev/canjs/node_modules/system-parse-amd/system-parse-amd.js:354:12
```

`socket.io-client/dist/socket.io` is a webpack built UMD module.  It looks something like:

```
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else if(typeof exports === 'object')
		exports["io"] = factory();
	else
		root["io"] = factory();
})(this, function() {
return /******/ (function(modules) { // webpackBootstrap
/******/ 	// The module cache
/******/ 	var installedModules = {};

/******/ 	// The require function
/******/ 	function __webpack_require__(moduleId) {

/******/ 		// Check if module is in cache
/******/ 		if(installedModules[moduleId])
/******/ 			return installedModules[moduleId].exports;

/******/ 		// Create a new module (and put it into the cache)
/******/ 		var module = installedModules[moduleId] = {
/******/ 			exports: {},
/******/ 			id: moduleId,
/******/ 			loaded: false
/******/ 		};

/******/ 		// Execute the module function
/******/ 		modules[moduleId].call(module.exports, module, module.exports, __webpack_require__);

/******/ 		// Flag the module as loaded
/******/ 		module.loaded = true;

/******/ 		// Return the exports of the module
/******/ 		return module.exports;
/******/ 	}


/******/ 	// expose the modules object (__webpack_modules__)
[...]

PR #580, opened 2017-01-11
Breaking test with multiple `define` calls.  

justinbmeyer:
Adds a test that shows a umd build with nested define calls breaks for #579.  This also lints a bit of the test code.

The `main` problem is this second `define` in _umd-ish.js_ that looks like:

```
	  if (isLoader) {
	    define(function () {
	      return JSON3;
	    });
	  }
```

I'm not sure what the best solution is without doing a lot of code analysis.  

My thought is that we could still give people a warning, but choose the `define` call with the most deps ... or the one that is the "least" nested as UMD factory functions will typically call `define` within only one other `function{ .... }`.

### DIFF
--- a/lib/graph/make_graph.js
+++ b/lib/graph/make_graph.js
@@ -3,7 +3,7 @@ var trace = require("../trace");
 var steal = require("steal");
 var _ = require("lodash");
 var clone = _.cloneDeep;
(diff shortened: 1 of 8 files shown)


justinbmeyer:
Hot damn!

PR #584, opened 2017-01-16
Test conditional builds using `~` lookup scheme

m-mujica:
https://github.com/stealjs/steal-conditional/issues/31


matthewp:
LGTM 👍

PR #589, opened 2017-01-24
Support JS minification through custom function

### LINKED ISSUE #524: Support JS minification through a custom function
This is a follow up on https://github.com/stealjs/steal-tools/issues/503, instead of supporting every minifier we are going to add support for a `minify` function:

`minify: function(js) { return myFancyMinifier(js); }`

Also, we are going to minify at the bundle level instead of that the module level

m-mujica:
Closes #524 

I tried this change using the custom build script below and it worked as expected!

The issue I see in this example is that babili gets the code after it has been transpiled, so we are not taking advantage of babili's Babili ES2015+ goodness 

```javascript
var path = require('path');
var babel = require('babel-core');
var stealTools = require('steal-tools');
var babiliPreset = require('babel-preset-babili');

var config = {
  config: path.join(__dirname, 'package.json!npm')
};

var options = {
  minify: function(source) {
    return babel.transform(source.code, {
      presets: [ babiliPreset ],
      sourceMaps: false,
      babelrc: false
    });
  }
};

stealTools.build(config, options)
  .then(function() {
    console.log('DONE!');
  })
  .catch(function() {
    console.log('FAILED: ', arguments[0]);
  });
```

### DIFF
--- a/test/multibuild_test.js
+++ b/test/multibuild_test.js
@@ -2229,7 +2229,7 @@ describe("multi build", function(){
 			}).then(function(){
 				open("test/umd/prod.html", function(browser, close){
 					find(browser, "umdExport", function(umdExport){
-						assert.equal(typeof umdExport, "function", "got umdExport");
+                        assert.equal(typeof umdExport, "function", "got umdExport");
 						close();
 					}, close);
 				}, done);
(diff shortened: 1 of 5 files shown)


test/multibuild_test.js
```diff
@@ -2229,13 +2229,40 @@ describe("multi build", function(){
...
-						assert.equal(typeof umdExport, "function", "got umdExport");
+                        assert.equal(typeof umdExport, "function", "got umdExport");
 						close();
 					}, close);
 				}, done);
 			}).catch(done);
 		});
 	});
 
+	it("supports passing custom minify funcions", function() {
```
m-mujica: ugh, will fix the typo later.

matthewp:
👍 LGTM

Needs docs, here http://stealjs.com/docs/steal-tools.BuildOptions.html

m-mujica:
docs added:

screen shot

PR #592, opened 2017-01-27
Fix default minify value for stealTools.build

### LINKED ISSUE #591: Minify doesn't work anymore with 1.1.0
I understand there are some change with minify.
But the default behaviour doesn't work anymore.
steal.production.js gets minified but bundles/main.js & bundles/main.css are not minified.

This is our build.js:
```

var buildPromise = stealTools.build({
    main    : "main",
}, {
  	bundleAssets	: false
});
```

Even with explicitly adding `minify: true` the bundle files are not minified.

m-mujica:
Closes #591

### DIFF
--- a/lib/build/multi.js
+++ b/lib/build/multi.js
@@ -79,6 +79,9 @@ module.exports = function(config, options){
 		return continuousBuild(config, options);
 
 	} else {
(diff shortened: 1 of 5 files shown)


matthewp:
👍

PR #599, opened 2017-02-04
update a package because of its dependency glob wich inhired minimatc…

frank-dspeed:
…h less then 3

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "steal-parse-amd": "^1.0.0",
     "through2": "^2.0.0",
     "tmp": "0.0.31",
-    "traceur": "0.0.91",
+    "traceur": "0.0.111",
     "transpile": "^1.0.0",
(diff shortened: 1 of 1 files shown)


matthewp:
Looks like it breaks tests, did they change the format maybe?

pYr0x:
`transpile`  `2´ will break even more tests :(

frank-dspeed:
Ok Nice looks like we know now what to do till next month lol :)

matthewp:
Try rebasing with master now. An upstream bug in steal was causing tests to break.

frank-dspeed:
hmm i am wondering the tests are showing traceur@0.0.91 should be traceur@0.0.111

matthewp:
@m-mujica Can you look into why upgrading traceur breaks the tests?

m-mujica:
Closing in favor of #627

PR #612, opened 2017-02-15
Basics of the development bundles feature

m-mujica:
- [x] Create a JS dev bundle 
- [x] Create a CSS dev bundle
- [x] Expose bundle command as `stealTools.bundle`
- [x] Allow filter modules using a glob pattern (default to `node_modules/**/*`)
- [x] Allow user to change destination folder (through `options.dest`)
- [x] Add plugins (e.g `done-autorender`) to dev bundle
- [x] Add `steal-tools bundle` command to CLI

```javascript

var config = { config: "package.json!npm" };

var options = {
  dest: "",                   // writes bundle to project root folder by default 
  filter: "node_modules/**/*" // default
};

stealTools.bundle(config, options).then(() => console.log("DONE!"))
```

While the option to load the dev bundle automatically is added, a workaround is to map `@dev` before the steal script as follows:

```html
<script>
  steal = {
    map: {
      "@dev": "dev-bundle"
    }
  };
</script>

<script src="node_modules/steal/steal.js"></script>
```

Place my order branch where I'm playing with this https://github.com/donejs/place-my-order/pull/138


lib/build/bundle.js
```diff
@@ -0,0 +1,63 @@
+var assign = require("lodash").assign;
+var minify = require("../stream/minify");
+var devBundle = require("../stream/dev_bundle");
```
matthewp: I think this should be renamed to `makeBundleStream` or something like that. The "dev" part of this feature should be pushed up into the cli.  The `--dev` flag is a special flag that expands to `filter: node_modules/**/*` or whatever we decide on.

That's the *only* place the word *dev* should be, in my opinion. This feature should be useable for creating other types of bundles, like bundles of just your own dependencies, perhaps.

lib/build/bundle.js
```diff
@@ -0,0 +1,63 @@
...
+var transpile = require("../stream/transpile");
+var concat = require("../bundle/concat_stream");
+var filterDevBundleGraph = require("../stream/filter_dev_bundle_graph");
```
matthewp: Same as above, `filterBundleGraph` instead of filterDevBundleGraph.

lib/stream/dev_bundle.js
```diff
@@ -0,0 +1,95 @@
...
+	splitMainBundles.filter(isJSBundle).forEach(function(bundle) {
+		var unshift = [].unshift;
+		unshift.apply(bundle.nodes, configGraph);
+
+		// Make config JS code so System knows where to look for bundles.
+		var configNode = makeBundlesConfig(
+			splitMainBundles,
+			configuration,
+			bundle,
+			{ excludedBundles: {} }
```
matthewp: I think this arg is optional? Might check, if so you can omit this line.

lib/build/bundle.js
m-mujica: @matthewp ok, I'll change this.... if `filter` is set to `node_modules/**/*` when `--dev` is passed,  should we throw if no `filter` value is passed for those using the API instead of the CLI?

matthewp: I don't think you should throw, it would just create a bundle of your entire graph, I think.

lib/stream/dev_bundle.js
m-mujica: I'd have to pass an empty object because of https://github.com/stealjs/steal-tools/blob/master/lib/bundle/make_bundles_config.js#L12.... I'll change the module to make `options` optional and get rid of this line,

lib/build/bundle.js
m-mujica: the filtered stream would be empty since `minimatch(anything, "");` always returns false.... Instead of using an empty string as default, I'll use the match anything blog pattern ("**") and that would indeed create a bundle of the entire graph.

m-mujica:
Just pushed a workaround to prevent the npm extension to load the package.json files.... I need to get a patched release of steal-npm/steal to get it working though.

matthewp:
@m-mujica you can do a patch release of steal if you need it.

matthewp:
Let's keep this in a branch. Can do a prerelease from within this branch. Want to avoid merging until the kinks are worked out.

m-mujica:
TL:DR; PMO loads ~30% faster and finishes requests 36% faster with dev bundles....

|  **Place My Order (without dev bundles)** | **Load Time (ms)** | **Finish Time (s)** |
|  ------ | ------ | ------ |
|   | 236 | 2.19 |
|   | 192 | 2.2 |
|   | 218 | 2.19 |
|   | 195 | 2.15 |
|   | 184 | 2.19 |
|   | 196 | 2.16 |
|   | 186 | 2.25 |
|   | 197 | 2.2 |
|   | 224 | 2.21 |
|   | 210 | 2.17 |
|   | 199 | 2.22 |
|   | 214 | 2.29 |
|   | **204.3** | **2.2** |


|  **Place My Order (with dev bundle)** | **Load Time (ms)** | **Finish time (s)** |
|  ------ | ------ | ------ |
|   | 168 | 1.59 |
|   | 175 | 1.43 |
|   | 148 | 1.43 |
|   | 186 | 1.41 |
|   | 140 | 1.39 |
|   | 138 | 1.4 |
|   | 132 | 1.34 |
|   | 137 | 1.41 |
|   | 112 | 1.41 |
|   | 137 | 1.39 |
|   | 127 | 1.37 |
|   | 141 | 1.42 |
|   | 125 | 1.38 |
|   | **143.5** | **1.4** |
|   |  |  |
|   |  |  |
|   | 0.29760156632403334 | 0.3636363636363637 |

m-mujica:
For bitballs, load time is roughly the same with dev-bundles, but it finishes all requests 25% faster when the dev bundle is loaded.

|  **BITBALLS (without dev bundles)** | **Load time (ms)** | **Finish time (s)** |
|  ------ | ------ | ------ |
|   | 124 | 2.28 |
|   | 139 | 2.16 |
|   | 137 | 2.22 |
|   | 135 | 2.19 |
|   | 139 | 2.2 |
|   | 138 | 2.21 |
|   | 129 | 2.14 |
|   | 137 | 2.23 |
|   | 130 | 2.18 |
|   | 136 | 2.16 |
|   | 143 | 2.22 |
|   | 135 | 2.18 |
|   | 141 | 2.18 |
|   | 143 | 2.19 |
|   | 137 | 2.18 |
|   | **136.2** | **2.2** |

|  **BITBALLS (with dev bundle)** | **Load time (ms)** | **Finish time (s)** |
|  ------ | ------ | ------ |
|   | 135 | 1.62 |
|   | 133 | 1.62 |
|   | 139 | 1.61 |
|   | 136 | 1.63 |
|   | 142 | 1.63 |
|   | 128 | 1.61 |
|   | 146 | 1.57 |
|   | 127 | 1.62 |
|   | 133 | 1.66 |
|   | 138 | 1.63 |
|   | 136 | 1.64 |
|   | 139 | 1.63 |
|   | 132 | 1.61 |
|   | 135 | 1.65 |
|   | 139 | 1.64 |
|   | **135.9** | **1.6** |

m-mujica:
Doing some profiling on bitballs, I found that the babel transpile is taking a good chunk of cpu time

screen shot 2017-02-23 at 09 18 42

... loading babel/transpiling in browser is slowing things down in dev mode (some babel calls take 10%+ CPU time and loading the babel extension takes are 2ms)

m-mujica:
screen shot 2017-03-01 at 16 50 07

doc/bundle.md
```diff
@@ -0,0 +1,119 @@
...
+dependencies in the `node_modules` folder.
+
+    var path = require("path");
+    var stealTools = require("steal-tools");
+
+    var promise = stealTools.bundle({
+      config: path.join(__dirname, "package.json!npm")
+    }, {
+      filter: "node_modules/**/*"
+    });
```
matthewp: You might want to use backticks for these codeblocks. At least GitHub has trouble thinking that the asterisks are not for bolding...

lib/stream/make_bundle.js
```diff
@@ -0,0 +1,140 @@
...
+		try {
+			var result = bundle(data);
+			done(null, result);
+		} catch(err) {
+			done(err);
+		}
+	});
+};
+
+function makeAddNpmPackagesNode(npmContext) {
```
matthewp: I think this code (and the code that calls it) should go in lib/bundle in its own module. We have other stuff like this, like [add_steal](https://github.com/stealjs/steal-tools/blob/master/lib/bundle/add_steal.js). This should follow that convention where it gets a bundle and unshifts itself into the top.

lib/stream/make_bundle.js
```diff
@@ -0,0 +1,140 @@
...
+		}
+	});
+
+	return _.assign({}, data, {
+		bundles: splitMainBundles
+	});
+}
+
+function isJSBundle(bundle) {
+	return bundle.buildType !== "css";
```
matthewp: This isn't necessarily true, there can be `buildType === 'html'` and others. I think you want to check `== null || === 'js'`.

PR #614, opened 2017-02-23
[DO NOT MERGE] Test steal-tools against steal#master

m-mujica:


m-mujica:
Arg,

```
 1) multi build Source Maps strip sourcemapping URL in all files:
      AssertionError: foobar.js.map should not be found
      + expected - actual
      -false
      +true
      
      at test/multibuild_test.js:2199:13
      at tryCatchReject (node_modules/steal/main.js:4891:30)
      at runContinuation1 (node_modules/steal/main.js:4850:4)
      at Fulfilled.when (node_modules/steal/main.js:4638:4)
      at Thenable.Pending.run (node_modules/steal/main.js:4529:13)
      at Scheduler._drain (node_modules/steal/main.js:3805:19)
      at Scheduler.drain (node_modules/steal/main.js:3770:9)
```

getting the same error locally, I don't think it's related to the loader change, but need to investigate.

m-mujica:
The test was failing because of a change in the Promise shim in `when.js`, in current master when.js was locked down to its previous version so tests should pass fine.

The test error message got me thinking it was something else :'( https://github.com/stealjs/steal-tools/pull/615

m-mujica:
tests pass as expecting, closing this.....

PR #619, opened 2017-03-02
Allow steal-tools to build arbitrary buildTypes

### LINKED ISSUE #586: HTML Build Type
steal-tools supports 2 build types currently, js and css. The css build type is marked by steal-css so that its resources get packed together into css bundles. 

I would like to do the same thing for HTML, so that a plugin for loading HTML could mark its resources as the `html` build type and steal-tools would therefore create separate HTML bundles.

matthewp:
This allows steal-tools to build arbitrary buildTypes. Basically it just
doesn't try to minify buildTypes it doesn't understand or add the
moduleName comment.

Fixes #586


m-mujica:
q4c7g8c

PR #621, opened 2017-03-06
Run tests against steal-conditional branch

m-mujica:


matthewp:
@m-mujica I think this can be closed, right?

Issue #630, opened 2017-03-14
react production build using steal-tools


reinaldoarrosi:
I am using the example available in https://github.com/stealjs/react-example to test stealjs.
The example works fine when following the steps on readme, then I tried to create a production build following these steps:

1. set NODE_ENV='production'
2. run steal-tools build
3. replace steal.js in index.html for ./dist/steal.production.js
4. running the app

The build worked but the following warning is shown when the app runs:

```
Warning: It looks like you're using a minified copy of the development build of React. When deploying React apps to production, make sure to use the production build which skips development warnings and is faster. See https://fb.me/react-minification for more details.
```

It seems that "steal-tools build" is minifying react but it is not removing code that is only used in development. According to react docs setting "NODE_ENV=production" should result in a version of react that does not contain development-only code.

Is this a bug or am I missing some configuration?

phillipskevin:
You can use the minified versions of React during the build by adding this to the `steal` section of your package.json:
```
    "envs": {
      "build-development": {
        "paths": {
          "react": "node_modules/react/dist/react.min.js",
          "react-dom": "node_modules/react-dom/dist/react-dom.min.js"
        }
      }
```

Someone else might have a better option, but this way worked when I tried it out.

matthewp:
Sounds like something that should go in the react-example docs. @reinaldoarrosi want to open an issue there? Or even better, a PR that adds this info to the docs.

reinaldoarrosi:
I've ended up customizing the minify function like this:
```javascript
var envify = require('loose-envify/replace'); // from https://www.npmjs.com/package/loose-envify
var uglify = require('uglify-js'); // from https://www.npmjs.com/package/uglify-js

...

stealtools.build(
{ 
    config: 'package.json!npm' 
}, 
{ 
    minify: function(source, options) {
        var envified = envify(source.code, [process.env]);
        var minified = uglify.minify(envified, { fromString: true });

        source.code = minified.code;
        return source;
    }
}
);
```

I've noticed steal-tools default minify function already uses uglify-js. Shouldn't it also use envify (like I did in my code)?
I think it would benefit not only react, but also any other packages that check for "process.env" since dead-code removal would be able to remove even more code.

matthewp:
@reinaldoarrosi I'm not familiar with envify, maybe it's something we should do. Want to open another issue to discuss what it is?

PR #632, opened 2017-03-15
support for envify in the default minify function

reinaldoarrosi:
This is the result of discussion in https://github.com/stealjs/steal-tools/issues/630 (specifically this comment https://github.com/stealjs/steal-tools/issues/630#issuecomment-286754274)

loose-envify (derived from the envify plugin for browserify) replaces Node-style environment variables with plain strings. 

For example, given that the environment variable NODE_ENV='dev' the following code
```javascript
if(process.env.NODE_ENV === 'prod') { /* run production-only code */ }
```

will be transformed to
```javascript
if('dev' === 'prod') { /* run production-only code */ }
```

This will then allow uglify-js dead code removal algorithm to remove this if statement since 'dev' === 'prod' will always be false. If NODE_ENV was set to 'prod' then the code would not be removed.

In this pull-request I've also added a new option "envify" which defaults to false and controls whether envify will or will not be called during minification.

### DIFF
--- a/lib/buildTypes/minifyJS.js
+++ b/lib/buildTypes/minifyJS.js
@@ -7,8 +7,9 @@ module.exports = function(source, options) {
 };
 
 function uglify(source, options) {
+	var envify = require("loose-envify/replace");
 	var UglifyJS = require("uglify-js");
-
+	
 	var code = source.code;
 	var opts = (options != null) ? options.uglifyOptions : {};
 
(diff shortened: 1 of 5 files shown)


matthewp:
Wow, totally cool of you to submit a PR for this. Even included a test! I'm impressed.  I'm on vacation currently, so probably won't be able to get to this until next tuesday. Looks good though!

matthewp:
Thanks for this!

PR #634, opened 2017-03-15
Test build with babel-standalone decorator plugin

m-mujica:
stealjs/steal/issues/1039


m-mujica:
Merging this one since it's not changing code, only adding a test.

PR #640, opened 2017-03-24
Support environment dependant babel plugins

### LINKED ISSUE #639: Make sure babel env plugins work
steal/issues/1102

m-mujica:
Closes #639

steal-tools will collect all babel plugins, and then it will:

- Add the built in plugins (names), even those in `env` if the babel environment matches 
- Collect all non-builtins that match the babel env, `require` them and push their functions into the plugins array passed to `transpile`, this way there is no need to register the custom plugins (names).


package.json
```diff
@@ -23,7 +23,7 @@
     "moment": "^2.10.2",
     "multimatch": "^2.1.0",
     "pdenodeify": "^0.1.0",
-    "steal": "^1.3.0",
+    "steal": "stealjs/steal#master",
```
m-mujica: I guess we need a release before we can merge this PR.

PR #646, opened 2017-04-03
Add failing test for steal/issues/952

m-mujica:


m-mujica:
The fix for this test landed on steal@1.4.1, will merge away captain!

PR #673, opened 2017-04-24
Disable clean-css import inlining 

### LINKED ISSUE #487: css @import will be removed by clean-css
if you have a stylesheet like:

**bar.css**

```
@import url("foo.css");
```

Make a build with minification, `clean-css` remove the `@import` statement. but does not include the **foo.css**. 

**solutions**
- set `processImport` default to `false` in `cleanCSSOptions`. the `@import` statement will not be touched. but you have also to set `bundleAssets` to `true`
- try to make import rules working with `clean-css`. maybe we have to set some options e.g. `root`

@matthewp what solution do you prefer

m-mujica:
- Disable clean-css import inlining
- Remove unused steal-css dependency
- Use clean-css promise based interface
- Refactor write_source to use a promise chain (such wow, much edit, very chainy)
- Split `writeStream` to handle errors coming from bundle's write_source stream  
- Add test to make sure steal-tools does not swallow clean-css errors

Closes #487

### DIFF
--- a/lib/build/multi.js
+++ b/lib/build/multi.js
@@ -100,10 +100,10 @@ module.exports = function(config, options){
 
 		// Return a Promise that will resolve after bundles have been written;
 		return new Promise(function(resolve, reject){
+			var bundleWriteStream = concatStream.pipe(createWriteStream());
+
 			// Pipe the build result into a write stream.
-			var writeStream = concatStream
-				.pipe(createWriteStream())
-				.pipe(stealWriteStream());
+			var writeStream = bundleWriteStream.pipe(stealWriteStream());
 
 			writeStream.on("data", function(data){
 				this.end();
@@ -126,6 +126,7 @@ module.exports = function(config, options){
 			  transpileStream,
 			  minifyStream,
 			  buildStream,
+		      bundleWriteStream,
 			  writeStream
 			].forEach(function(stream){
 				stream.on("error", reject);
--- a/test/multibuild_test.js
+++ b/test/multibuild_test.js
@@ -2072,4 +2072,29 @@ describe("multi build", function(){
 				});
 			});
 	});
+
+	it("does not swallow clean-css errors", function(done) {
+		var base = path.join(__dirname, "broken_css_build");
+
+		asap(rmdir)(path.join(base, "dist"))
+			.then(function() {
+				return multiBuild({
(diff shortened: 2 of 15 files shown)


lib/build/multi.js
```diff
@@ -100,10 +100,10 @@ module.exports = function(config, options){
 
 		// Return a Promise that will resolve after bundles have been written;
 		return new Promise(function(resolve, reject){
+			var bundleWriteStream = concatStream.pipe(createWriteStream());
```
m-mujica: I hate that `.pipe` does not propagate errors </rant>

matthewp: Yeah, weird huh.

m-mujica:
@matthewp 👍  👎  😕  ?

test/multibuild_test.js
```diff
@@ -2072,4 +2072,49 @@ describe("multi build", function(){
...
+			.then(function() {
+				return multiBuild({
+					main: "main",
+					config: path.join(base, "stealconfig.js")
+				}, {
+					quiet: true
+				});
+			})
+			.then(function() {
+				var source = fs.readFileSync(path.join(base, "dist", "bundles",
```
pYr0x: is `foo.css` also copyed to the dist folder?

m-mujica: There is no need to copy it over; steal-tools rewrites the `@import` urls so they work from within the dist folder..... Something similar to what `clean-css` with the `rebaseTo` property.

matthewp:
LGTM

test/multibuild_test.js
pYr0x: Steal tools will rewrite the URL to the location of 'foo.css' ? But sometimes you only want to publish the dist folder. 
Or what happened if you import e.g. Bootstrap from the node_modules folder...? 
For JavaScript files you don't need the node_modules folder built. But if you use a CSS from there you will need it.

m-mujica: don't use `@import` then, it will cost you a request in production, load it through steal.... If you think this is a problem please file and issue and we'll discuss it there.

matthewp: Yeah, we can possibly inline the styles. There are arguments in favor of that.  This PR just preserves the old behavior from 0.16.x.

PR #685, opened 2017-04-27
Make sure conditional substitution builds correctly

m-mujica:
Part of stealjs/steal-conditional#38

The idea is to merge https://github.com/stealjs/steal-conditional/pull/40 first, patch release and update this PR.


m-mujica:
I'm going to merge this one since it's only adding (and refactoring) tests.

PR #688, opened 2017-04-28
Remove engine-dependencies package: ~POINTS

### LINKED ISSUE #687: Deprecate engine-dependencies
steal-tool is tested against node 4, 6 and current.

it'd be nice to get rid of the engine-dependencies package/configuration/script and make zombie a regular dev dependency.

m-mujica:
Closes #687


matthewp:
:D

PR #693, opened 2017-05-01
Prevent clean-css to rebase urls

### LINKED ISSUE #689: CleanCSS incorrectly updates `url()`s when using bundles
CleanCSS will update `url()`s under certain [conditions](https://github.com/jakubpawlowicz/clean-css#how-to-rebase-relative-image-urls) in CSS when [minifying](https://github.com/stealjs/steal-tools/blob/master/lib/buildTypes/minifyCSS.js#L13-L14). These `url()`s are already updated prior to bundling, so this can cause URLs to be broken. I believe we should set `rebase: false` in the [`cleanCSSOptions`](https://github.com/stealjs/steal-tools/blob/master/lib/assign_default_options.js#L14-L16).

m-mujica:
Closes #689

### DIFF
--- a/lib/assign_default_options.js
+++ b/lib/assign_default_options.js
@@ -12,6 +12,7 @@ module.exports = function(config, options){
 		bundleSteal: false,
 		uglifyOptions: {},
 		cleanCSSOptions: {
+			rebase: false,
 			inline: ["none"]
 		},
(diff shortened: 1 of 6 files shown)


m-mujica:
@matthewp can I merge this one?

Issue #666, opened 2017-04-20
Create basic slim bundler api


matthewp:
This is an internal api that will be used by `stealTools.build`. The name doesn't matter for now, maybe "optimize"? Since this is like an optimization.

# optimize(options)

```js
optimize({
  // An array of steal-tools bundle objects
  bundles: [
    {nodes: [], buildType: 'js'}
  ],

  // Don't worry about the other options for now but this is the idea
  plugins: true,
  otherOptions: true
});
```

With the return value I'm guessing being the bundles array with any shim stuff added to the beginning and/or ends.

m-mujica:
@matthewp in this example https://github.com/stealjs/steal-tools/issues/663#issuecomment-300544185

The generated things looks like this:

```js
(function(modules) {
  ....
})
```

The IIFE with the "loader" logic and then the actual function call with the nodes as an array.

```js
([
     function() {},
     function() {},
     function() {},
])
```

How should we arrange the nodes in the bundle to generate this output? I mean, the node can't contain partial code, so I could unshift the IIFE with the loader code but what about the actual function call?

```js
{ bundle: [
     { /* IIFE loader code */ },
     { /* bundle node */   },
     { /* bundle node */   },
     { /* bundle node */   },
     { /* bundle node */   }
  ]
}
```

Should transpile be responsible of taking the first node and then create the arguments list based on the following nodes in the bundle?

m-mujica:
actually transpile can't do that since transpile gets acts on every node. It's a steal-tools thing... At the concatenate bundle level maybe.

matthewp:
I would think that in "slim mode" it would have a different concat bundle module since, as you said, it does something different from what the regular concatenate would do.  Otherwise, I think you are right, it would just unshift the loader into the bundle array.

Issue #663, opened 2017-04-20
Make the slim loader basic shim


matthewp:
The basic shim should be able to execute modules that conform to the AMD commonJS format like:

```js
define('bar', function(module, require){
  module.exports = require('bar');
})
```

But note that we are using numbers instead of names and we don't want to use `define` or `require` to prevent any format detection if loaded through another loader (like Steal or webpack).

Also we need to make this work with `script=async` (worry about this in another issue) so we have to account for the loader being executed **after** bundles.

So let's go with this for now (might change it before release):

```js
(_steal=_steal||[]).push([
  [1, function(stealModule, stealRequire){
    stealModule.exports = stealRequire(2);
  }],
  [2, function(stealModule, stealRequire){
    stealModule.exports = "bar";
  }]
]);
```

The basic shim will be attached to the top of this bundle by default. It will also be possible to put the shim into a separate bundle (so you can load multiple bundle in async scripts as the same time without them all having the shim).  but don't worry about that for now.

So the shim needs to be able to execute the above bundle, and it needs to be about the same size as webpack's basic shim (hopefully smaller!).  

See https://github.com/matthewp/webpack-research under the "single" folder which contains their code for loading a single bundle like this.

m-mujica:
The app's code:

```js
// foo.js
module.exports = "foo";
```

```js
// bar.js
module.exports = "bar";
```

```js
// index.js
var foo = require("foo");
var bar = require("bar");

window.foo = foo;
window.bar = bar;
```

The bundle code: 

```js
(function(modules) {
  var cache = {};

  function stealRequire(moduleId) {
    if (cache[moduleId]) {
      return cache[moduleId];
    }

    var mod = cache[moduleId] = {
      exports: {}
    };

    modules[moduleId].call(mod.exports, stealRequire, mod.exports, mod);
    return mod.exports;
  }

  // import the main module
  stealRequire(2);
})([

  /*foo[0]*/
  function(stealRequire, stealExports, module) {
    module.exports = "foo";
  },

  /*bar[1]*/
  function(stealRequire, stealExports, module) {
    module.exports = "bar";
  },

  /*index[2]*/
  function(stealRequire, stealExports, module) {
    var foo = stealRequire(0);
    var bar = stealRequire(1);
    window.foo = foo;
    window.bar = bar;
  }
]);
```

matthewp:
@m-mujica Submit PRs to minor branch I guess (if it doesn't exist create one off of minor).

[...]

Issue #668, opened 2017-04-20
Enable slim bundles to be loaded with async scripts


matthewp:
This should come after https://github.com/stealjs/steal-tools/issues/667

A big win for this feature is that bundles can be loaded asynchronously.

```html
<script src="main-bundle.js" async></script>
<script src="bundle-a" async></script>
<script src="bundle-b" async></script>
```

Where **main-bundle.js** contains the loader. It doesn't matter if **bundle-a** or **bundle-b** execute first, because all they do is push a value into an array like:

```js
(_steal=_steal||[]).push([
  [1, function(stealModule, stealRequire){
    stealModule.exports = stealRequire(2);
  }],
  [2, function(stealModule, stealRequire){
    stealModule.exports = "bar";
  }]
]);
```

The loader will know when all necessary bundles have loaded and will execute the modules only when all are complete.

m-mujica:
@matthewp does `main-bundle` imports any of the stuff inside either `bundle-a` or `bundle-b` ? 

Because in that case, the bundles will be part of the `main-bundle` dependency graph and will be bundled into `main-bundle`, right?

I'm trying to set up a simple app with this use case in mind.

m-mujica:
and if `main-bundle` does not need any of the stuff in `bundle-a` or `bundle-b`, why to load them? To preload the "next" page code or something?

matthewp:
No, I don't think `main-bundle` will ever import stuff in `bundle-a` or `bundle-b`.

Yes, the idea is that if you have a `main-bundle` and a `home-bundle` and you are visiting the homepage you can add both script tags have them download in parallel. This prevents the waterfall effect that steal apps have today.

m-mujica:
Closed by https://github.com/stealjs/steal-tools/pull/715

PR #702, opened 2017-05-11
Get basics working

m-mujica:
Issues

- https://github.com/stealjs/steal-tools/issues/663
- https://github.com/stealjs/steal-tools/issues/666

@matthewp  thoughts? Next I'm going to try adding some deps, and then maybe progressive loading 🎸

### DIFF
--- a/lib/build/slim.js
+++ b/lib/build/slim.js
@@ -1,0 +1,33 @@
+var pump = require("pump");
+var assignDefaultOptions = require("../assign_default_options");
+
+var streams = {
+	bundle: require("../stream/bundle"),
+	transpile: require("../stream/transpile"),
+	concat: require("../bundle/concat_slim_bundle"),
+	write: require("../bundle/write_bundles").createWriteStream,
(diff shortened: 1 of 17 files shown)


matthewp:
I took a look at the code. No issues from me, looks good.  I think the progressive loading case is going to require changing the shim but I think it's ok if there's a different shim for single vs multi case like it appears you are doing here.

I think once you add runtime plugins (css) you'll want to share code between the multi with no-runtime and multi with runtime cases, but can cross that bridge later.

I think progressive loading next is a good idea, but keep an eye on #668 as well (async scripts) as that is closely related.

m-mujica:
@matthewp just updated this PR to add support for the progressive/async version of the shim.

I'm thinking that maybe I should work on the module ids before moving to transpile, since transpile works at the node level. 

What do you think?

matthewp:
Yeah, whatever you think is best, I just meant it might be good to work towards a full example before moving on to plugins.

PR #705, opened 2017-05-15
Fix steal-tools bundle docs

m-mujica:
- Remove references to `deps-bundle-dest` and `dev-bundle-dest`; the attributes `dev-bundle` and `deps-bundle` can be passed a string to the path where the dev-bundle module is located.

- Adjust indentation of the html tags with multiple attribute to avoid horizontal scrolling in the generated code blocks.

screen shot 2017-05-15 at 08 35 17

Closes #1197


matthewp:
👍 closes https://github.com/stealjs/steal/issues/1197 actually

m-mujica:
@matthewp oops, let me amend the commit message.

PR #709, opened 2017-05-16
Set greenkeeper to ignore uglify-js

m-mujica:
steal@2 will probably drop uglify-js and use a es6 friendly minifier; uglify-js@3 has a breaking api and we are not planning to get it to work, yet...


m-mujica:
@matthewp can we merge this one? To prevent greenkeeper creating more PRs to update uglify-js

PR #722, opened 2017-06-02
Watch mode should not minify by default

### LINKED ISSUE #708: Watch mode isn't defaulting to minify false
It shouldn't minify by default. https://github.com/stealjs/steal-tools/blob/master/lib/build/continuous.js#L15

I think this is probably caused by the cli.

m-mujica:
Closes #708 

This one turned out to be a bit tricky, because you still want to minify if the flag was passed as in:

```
steal-tools build --watch --minify
```

but since we were using yargs default value, we could not tell whether minify was set by the user at the point were we were looking at `watch` value.

The solution was to stop using the default minify value only in the build command and pass options through a function that takes care of setting its value.

### DIFF
--- a/lib/cli/cmd_build.js
+++ b/lib/cli/cmd_build.js
@@ -1,8 +1,15 @@
 var winston = require("winston");
 var stealTools = require("../../index");
+var makeStealConfig = require("./make_steal_config");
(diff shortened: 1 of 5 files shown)


matthewp:
👍

PR #729, opened 2017-06-05
Watch mode should have quiet output by default

### LINKED ISSUE #707: Watch mode isn't defaulting to quiet output
It should be quiet by default: https://github.com/stealjs/steal-tools/blob/master/lib/build/continuous.js#L16

I think the cli code is forcing it to be true.

m-mujica:
Closes #707

negated-dependant properties like no-minify/minify or verbose/quiet make me want to use computes to keep them in sync :D


matthewp:
👍

PR #733, opened 2017-06-07
Use the slim wrapper only in JS bundles

### LINKED ISSUE #732: Use the slim wrapper only in JS bundles
Currently the logic that concatenates the slim bundles and wraps them with the slim array format does not check `load.metadata.buildType` causing css (and other stuff) to be incorrectly wrapped. 

```js
/*bundles/slim-css/index.css!*/
(__steal_bundles__ = window.__steal_bundles__ || []).push([
	5,
	"body {    color: blue;  }  "
]);
```

m-mujica:
Closes #732


matthewp:
LGTM 

giphy

PR #735, opened 2017-06-09
Provide a way for plugins to be written for slim 

m-mujica:
ue9rway

### DIFF
--- a/lib/stream/load_optimized_plugins.js
+++ b/lib/stream/load_optimized_plugins.js
@@ -1,0 +1,56 @@
+var keys = require("lodash/keys");
+var through = require("through2");
+var assign = require("lodash/assign");
+var denodeify = require("pdenodeify");
+var cloneDeep = require("lodash/cloneDeep");
+var readFile = denodeify(require("fs").readFile);
+
+module.exports = function() {
+	return through.obj(function(data, enc, next) {
+		loadPlugins(data)
+			.then(function(newGraph) {
+				next(null, assign(data, { graph: newGraph }));
+			})
+			.catch(next);
+	});
+};
+
+/**
+ * Replaces plugin's source with its `pluginBuilder` module
+ * @param {Object} data - The stream's data object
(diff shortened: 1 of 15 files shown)


m-mujica:
@matthewp this is a naive implementation but it works, I was able to get [steal-css/slim](https://github.com/stealjs/steal-css/blob/slim/slim.js) to be included in the build

ue9rway


I should be able to replicate the [build](https://github.com/m-mujica/slim-css/tree/master/out/bundles/slim-css) in my example tomorrow, need to get the config in place and the conditional in the require

matthewp:
👍 looks good to me.

m-mujica:
@matthewp I did quite a few changes here, mind taking a look again?

lib/stream/load_optimized_plugins.js
```diff
@@ -0,0 +1,56 @@
...
+		var pluginBuilder = node.value && node.value.pluginBuilder;
+
+		if (node.isPlugin && pluginBuilder) {
+			return loader
+				.normalize(pluginBuilder)
+				.then(function(name) {
+					return loader.locate({ name: name, metadata: {} });
+				})
+				.then(function(address) {
+					return Promise.all([address, readFile(address.replace("file:", ""))]);
```
matthewp: Wouldn't it be better to use `loader.fetch` here?

lib/stream/load_optimized_plugins.js
```diff
@@ -0,0 +1,64 @@
...
+						return Promise.all([
+							name,
+							loader.locate({ name: name, metadata: {} })
+						]);
+					})
+					// [ name, address ]
+					.then(function(data) {
+						return Promise.all([
+							data[1],
+							loader.fetch({ name: data[0], address: data[1], metadata: {} })
```
m-mujica: @matthewp done!

matthewp: 👍 merge then

PR #744, opened 2017-06-15
Detect code that cannot be used with slim

### LINKED ISSUE #743: Add ability to detect code that can't be used with slim loader
We need some method for detecting code that can't be used with the slim loader. Some examples today are:

* Code that uses `@loader` or `@steal`.
* Code that uses steal-conditional
* Code that relies on envs config for `window-production`.

We can detect these things and throw an appropriate message. Over time we want to remove these restrictions. First step would be getting infrastructure in place so we can write these checks and remove them when a proper solution is implemented.

m-mujica:
Closes #743 

Checking the graph was trickier than expected :/


matthewp:
I kind of feel like it would be better if this were broken up into some type of rules. Like where each rule is in a `rules/steal-conditional.js` and `rules/steal-and-loader-modules.js`, etc. Your way is more efficient since you only traverse the graph once, but I sort of feel doing it the other way would make it more readable and easier to remove rules. In the future if we make `@loader` work we would just delete the `rules/steal-and-loader-modules.js` file and that would be it.

What do you think?

m-mujica:
I think it's a good idea; on it!

m-mujica:
@matthewp how does it look now?

matthewp:
👍 looks good!

PR #745, opened 2017-06-16
Allow the loader to be put into its own bundle

### LINKED ISSUE #669: Allow the loader to be put into its own bundle
If the user has many bundles, by default the loader shim will be in the main bundle like:

```html
<script src="main-bundle.js" async></script>
<script src="bundle-a.js" async></script>
<script src="bundle-b.js" async></script>
```

However, you *might* want to put it into its own bundle (not sure why exactly) so we should support that:

```html
<script src="loader.js" async></script>
<script src="main-bundle.js" async></script>
<script src="bundle-a.js" async></script>
<script src="bundle-b.js" async></script>
```

Through some type of option. One possibility:

```js
optimize({
  bundles: [],

  splitLoader: true
})
```

m-mujica:
Closes #669 

giphy

### DIFF
--- a/test/slim/progressive/split.html
+++ b/test/slim/progressive/split.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<title></title>
+</head>
+<body>
+	<script src="./dist/bundles/loader.js" async></script>
+	<script src="./dist/bundles/main.js" async></script>
+	<script src="./dist/bundles/baz.js" async></script>
+</body>
+</html>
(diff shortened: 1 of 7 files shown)


test/slim/progressive/split.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<title></title>
+</head>
+<body>
+	<script src="./dist/bundles/loader.js" async></script>
+	<script src="./dist/bundles/main.js" async></script>
+	<script src="./dist/bundles/baz.js" async></script>
```
matthewp: This site is so beautiful

matthewp:
:+1

PR #749, opened 2017-06-19
Enable minification in the slim build stream

### LINKED ISSUE #748: Slim build minification is not working

m-mujica:
Closes #748 

@matthewp I had to push the minification down the stream right before the bundles are written out, the reason being "slim nodes" will be minified to empty source since the minifier sees them as unused array declarations.

The downside of this is that bundles are going to be computed with the raw source, what do you think?


matthewp:
Hm, I see why you did it this way.  Let's go with it for now and come up with a better solution in the future, since this only affects slim bundles and that's new.

PR #750, opened 2017-06-21
documented uglifyOptions with an example

ilyavf:
Some libraries may rely on function names (e.g. `bitcoinjs-lib` uses some type checking).
Documented the option to prevent uglifying function names.


matthewp:
👍 thanks!

PR #752, opened 2017-06-22
Return module exports from cache (instead of `module`)

### LINKED ISSUE #751: slim loader breaks when serving modules from cache
`stealRequire` returns an object and the slim loader is trying to call it as if it were a function when the module is served from the loader's cache.

m-mujica:
Closes #751

The weather widget is working with this version 😭 

google chrome


m-mujica:
merging this one 👀

PR #754, opened 2017-06-23
Remove flagged nodes from the slim build

### LINKED ISSUE #753: slim loader should remove flagged node from the plugins
The regular build does this in the `concat` stream; unfortunately the slim stream creates single node bundles before the `concat` stream does its thing.

Moving down the slim graph logic after concat breaks the logic to create the slim nodes.

m-mujica:
Closes #753 

Branched off the PR for #751 

The "puppies" widget build makes a lot more sense with this change, it seems the only missing thing is the GLOBAL->SLIM transpiler, I'll do that tomorrow.


matthewp:
Can the isExcludedFromBuild function be shared with the multibuild? I know there's a similar function there, is it different at all?

m-mujica:
@matthewp maybe, I thought about it, there is this logic https://github.com/stealjs/steal-tools/blob/master/lib/bundle/concat_source.js#L32 that made me think filtering first would break something.... But let me try sharing it and see if there is tests pass.

matthewp:
@m-mujica I just mean moving the function to a shared location, not changing the order in which you do things.

PR #757, opened 2017-06-24
Remove flagged nodes from the slim build

### LINKED ISSUE #753: slim loader should remove flagged node from the plugins

m-mujica:
Closes #753


m-mujica:
@matthewp finally got it working! 😭 

This is the PR in my fork of myhub, the biggest change is https://github.com/stealjs/myhub/pull/4/files#diff-7b9ea9fd8d2c98815d2c55a9bdc128efR19, I had to use the full ids there b/c the minimal loader does not do normalization on runtime.

In order to get the slim command to work in `myhub` I had to manually install `npm i stealjs/steal-css#slim`; changing the package.json to

```
"steal-css": "stealjs/steal-css#slim"
```

breaks the npm plugin.

I need to get a prerelease of steal-css#slim, but for some reason steal-css's build is broken, will look into that tomorrow

google chrome

.

matthewp:
Cool!  Can you create an issue about having to use the full module IDs? Maybe something to tackle in phase 2.

Also, when writing the guide maybe do a perf comparison. Not sure if it will be significant in this case or not.

PR #760, opened 2017-06-28
Add docs and a guide for steal-tools optimize builds

### LINKED ISSUE #718: Create a guide for the slim loader
We should create a guide on converting a project to use the slim loader. I think something like donejs-chat would be a good example as it *should* be usable with slim.

m-mujica:
Closes #718

### DIFF
--- a/doc/guides/optimized_builds.md
+++ b/doc/guides/optimized_builds.md
@@ -1,0 +1,105 @@
+@page StealJS.guides.optimized_builds Optimized Builds
+@parent StealJS.guides
+
+@body
+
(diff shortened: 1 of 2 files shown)


m-mujica:
@matthewp a couple of questions:

1) should we keep the guide in steal-tools or in steal?
2) In the guide there is a step about updating the dynamic module ids. We need to get myhub working with phase 1 slim, but it's awkward. Should we explain what the full module name is? how to do the translation manually? :/
3) About the perf comparison, what would you like to see?
4) Should we add a section in the guide to use `splitLoader` and show async loading

Any feedback is appreciated.

matthewp:
🚢 🇮🇹

PR #762, opened 2017-06-29
Slim guide revisions

### LINKED ISSUE #761: Slim guide revisions
From looking at the guide there were a few things I noticed, mostly around the comparison part:

For reference [here is the multibuild](https://user-images.githubusercontent.com/724877/27653018-0529f708-5bfb-11e7-91c7-27b988ef7b18.png) and [here is the slim build](https://user-images.githubusercontent.com/724877/27653129-5d2fb35c-5bfb-11e7-85fb-fa48f2a79e1b.png)

1. Progressive loading is part of the guide (we update the identifier) but this is not shown in the network tab. Why is that?
    * Worth noting that with the slim build we could have 2 different pages (one for each page), that way we can add the appropriate async tags to each. I think it's worth adding this to the guide since it's a major benefit of the slim build.
2. The multibuild version is getting 304 responses, so it's coming from the cache. Turn off caching so that we get accurate comparisons between the two.
3. I would notice the different in file size. The multibuild is 182k and the slim build is 88k according to the network tabs. But I bet that http-server doesn't gzip, you can use a tool like [gzip-size-cli](https://www.npmjs.com/package/gzip-size-cli) to get the gzip sizes.

m-mujica:
Closes #761

### DIFF
--- a/doc/guides/optimized_builds.md
+++ b/doc/guides/optimized_builds.md
@@ -3,7 +3,7 @@
 
 @body
 
-In 1.4.0 StealTools added a new **optimize** api which aims to replace [stealTools.build](steal-tools.build)
+In 1.4.0 StealTools added a new **optimize** API which aims to replace [stealTools.build](steal-tools.build)
 as the default way of creating a build of a module and all of its dependencies; builds created by `stealTools.optimize` are smaller and load faster.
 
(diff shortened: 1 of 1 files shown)


m-mujica:
@matthewp is this what you had in mind?

I found a bug with the async loading of the bundles, will take care of that tomorrow morning. And then maybe expand a little bit on the final section.

matthewp:
Yep, looks good.

m-mujica:
made a small clean-up, added another screenshot and verified the built page looks fine

google chrome

PR #764, opened 2017-06-29
Flag bundles as LOADED even if not required

### LINKED ISSUE #763: async progressive bundles not working
In myhub, when the progressive bundles are loaded with async script tags, clicking on the navigation links does not work.

m-mujica:
The slim loader was only flagging bundles as loaded if the bundles were already required; that caused lazy loaded bundles to not resolve correctly.

Closes #763


matthewp:
👍

PR #772, opened 2017-06-30
Fail the build if circular dependencies are found

### LINKED ISSUE #771: [SLIM] Add a check for circular dependencies
The issue https://github.com/stealjs/steal-tools/issues/742 is about adding support to the loader, for the next release we'll fail the build if there are circular dependencies.

m-mujica:
Closes #771


matthewp:
lgtm

PR #773, opened 2017-06-30
Add file size benchmarks

### LINKED ISSUE #664: Create file size benchmarks for common scenarios
The scenarios of the slim loader usage is:

1. A single bundle (see [this example](https://github.com/matthewp/webpack-research/tree/master/single)).
2. Multiple bundles, progressively loaded (see [this example](https://github.com/matthewp/webpack-research/tree/master/split)).
3. Single or multiple bundles using a plugin (there are no runtime plugins in webpack).

We should create some tests that have a basic example of each of these. The test will compare the size of a generated build to that of a snapshot. 

This way we can prevent accidental feature-creep affecting file size. That's not to say we won't ever increase the size of the shim, but when we do it will break a test and we'll have to explicitly make the decision to create a new snapshot.

m-mujica:
Closes #664

screen shot 2017-06-30 at 17 17 38

screen shot 2017-06-30 at 17 16 54

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "find-line-column": "^0.5.2",
     "fs-extra": "~3.0.0",
     "glob": "^7.1.1",
+    "gzip-size": "^3.0.0",
     "is-appveyor": "^1.0.0",
     "lodash": "^4.3.0",
     "loose-envify": "^1.3.1",
--- a/test/check_size_snapshot.js
+++ b/test/check_size_snapshot.js
@@ -1,11 +1,17 @@
 var path = require("path");
 var fs = require("fs-extra");
 var colors = require("colors");
+var range = require("lodash/range");
 var gzipSize = require("gzip-size");
 var denodeify = require("pdenodeify");
+var includes = require("lodash/includes");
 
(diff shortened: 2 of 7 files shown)


m-mujica:
@matthewp is this what you had in mind?

matthewp:
Yeah, looks good. Windows failing is probably a path issue, right?

matthewp:
We just need to remember to add a snapshot test for each new add-on we make.

Do you want to try and reduce the size here or in another issue?

m-mujica:
@matthewp yeah, different issue sounds good. I'll create the issue and work on it. 

Interesting the window bundles are bigger than the ones in MacOS, I guess b/c the module identifier in the config are different, should I add some kind of offset to the comparation? The biggest different seems to be 3b

m-mujica:
https://ci.appveyor.com/project/matthewp/steal-tools-427/build/1.0.1337#L1296

m-mujica:
The `offset` thing is not very elegant, but it did the trick. Thoughts?

matthewp:
Good enough I guess :shrug:

PR #777, opened 2017-07-06
Support ESM named imports

### LINKED ISSUE #776: [SLIM] esm named imports not working
Currently the loader _always_ tries to return `.default` when it sees the `__esModule` flag, this does not work for named imports.

**We need the ES6 -> SLIM transpiler to wrap the `stealRequire` call with the `stealRequire.esm` function only for default imports.**

Babel transpiles the following named import

```js
import { foo } from "./foo";
foo()
```

to 

```js
var _bar = require("./bar");

(0, _bar.foo)();
```

The slim version of this module should look like:

```js
[ 
  "foo", 
    var _bar = stealRequire.esm(stealRequire("bar"));
    (0, _bar.foo)();    
  }
]
```

where `stealRequire.esm` looks at the `__esModule` flag and returns `stealExports.default` instead `stealExports`.

This can be a little bit challenging since we rely on babel to do the transpilation and when we get the AMD output, we no longer know whether a default import was used.

m-mujica:
Closes #776

### DIFF
--- a/lib/node/slim/make_shim_template.js
+++ b/lib/node/slim/make_shim_template.js
@@ -24,7 +24,7 @@ module.exports = function(options) {
 
 			function stealRequire(moduleId) {
 				if (loadedModules[moduleId]) {
-					return stealRequire.esm(loadedModules[moduleId].exports);
+					return loadedModules[moduleId].exports;
 				}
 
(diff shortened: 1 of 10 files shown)


matthewp:
🚢 🇮🇹

PR #778, opened 2017-07-06
Support circular dependencies (slim loader)

### LINKED ISSUE #742: Circular dependencies in slim builds
We should figure out if this is something that works with webpack, and regardless see if this is something that can be added.  If it adds code (more than a couple of lines) it should be something that is added conditionally only when needed.

m-mujica:
Closes #742

In order to merge this one:

- [x] Merge and release https://github.com/stealjs/transpile/pull/97
- [x] Update package.json to use the `transpile` version released before
- [ ] do merge and close the issue


matthewp:
:D looks good to me!

PR #779, opened 2017-07-07
Add partial support for @loader module

### LINKED ISSUE #721: Provide the ability to use some of @loader in the slim loader such as envs configuration
@matthewp commented on [Thu Jun 01 2017](https://github.com/stealjs/steal-conditional/issues/44)

Modules that use slim loader like:

```js
var loader = require("@loader");

var url = loader.serviceBaseURL + "/api/todos";
```

Should not throw. I think the way to approach this is that steal-tools can provide a shim module for `@loader`. It can then attach the envs configuration to that loader object so that if it looks like:

```json
"envs": {
  "window-production": {
    "serviceBaseURL": "http://example.com"
  }
}
```

The pseudo-code for the loader module would be:

```
module.exports = {
  "serviceBaseURL": "http://example.com"
}
```

m-mujica:
Closes #721


matthewp:
Unfortunately the loader contains a lot of state that is specific to the loading process and not necessarily relevant for the client-side production use.  I guess you did it that way to make sure we get config whether it comes from the package.json or `steal.config` calls?  What you *could* do, instead, is have the build override `loader.config` and capture anything that comes in, copying the result. This way you would know that is meant for config use.

m-mujica:
This is the output of the `@loader` test sample app

```js
/*[slim-loader-config]*/
(function(global) {
	global.steal = global.steal || {};

	global.steal.map = {};
	global.steal.paths = { "4": "dist/bundles/main.js" };
	global.steal.bundles = { "2": 4, "3": 4 };
	global.steal.plugins = {};
})(window);

/*[slim-loader-shim]*/
(function(modules) {
	var modulesMap = {};
	var loadedModules = {};

	/*globals steal, __steal_bundles__*/

	function addModules(mods) {
		mods.forEach(function(m) {
			modulesMap[m[0]] = m[1];
		});
	}
	addModules(modules);

	function stealRequire(moduleId) {
		if (loadedModules[moduleId]) {
			return loadedModules[moduleId].exports;
		}

		var stealModule = (loadedModules[moduleId] = {
			exports: {}
		});

		modulesMap[moduleId].call(
			stealModule.exports,
			stealRequire,
			stealModule.exports,
			stealModule
		);

		return stealModule.exports;
	}

	// import the main module
	stealRequire(2);
})([
	[
		3,
			stealModule.exports = {
				logLevel: 3,
				meta: {},
				stealPath: "/Users/mmujica/Projects/steal-tools/node_modules/steal",
				main: "main",
				envs: { "window-production": { serviceBaseURL: "/api/production" } },
				serviceBaseURL: "/api/production"
			};
		}
	],
	[
		2,
			var loader = stealRequire(3);
			window.serviceBaseUrl = loader.serviceBaseURL;
		}
	]
]);
```

m-mujica:
@matthewp I had to remove the envs config check along with the `@loader` check; thing is we don't  fully support both things... Should we output a warning? I don't like warnings but....

matthewp:
@m-mujica Yeah, I think a warning is ok here. Just to explain that they might be using `@loader` in a way that won't work in production.

matthewp:
The only thing that gives me pause is `stealPath` having a fs path in it. The npm plugin used to contain filesystem paths and that was a problem for some people. I wonder if we should blacklist (yuck) a few properties that we know will contain fs paths but that people are unlikely to actually be using in production anyways. `stealPath` and (maybe) `paths` are the two that come to mind here...

m-mujica:
@matthewp I changed it to filter out those properties and added a warning. I guess we are not merging this one yet, right?

m-mujica:
well, 1.4.0 is out so I guess we could indeed merge it

matthewp:
Hey, yeah we can merge. We'll release tomorrow... not sure whether to do 1.5.0 or 1.4.x... will discuss tomorrow.

PR #783, opened 2017-07-10
Use non-normalize bundle names in slim config

### LINKED ISSUE #759: [SLIM] Support module identifiers in dynamic imports
Currently, you have to use the full normalized module name in dynamic imports, e.g:

```javascript
steal.import(`myhub@1.0.0#${hash}/${hash}`); // works!
```

instead of 

```javascript
steal.import(`~/${hash}/${hash}`);  // throws since the slim loader doesn't normalize 
steal.import(`myhub/${hash}/${hash}`); // same here
```

Dynamic module ids are resolved during runtime; the minimal loader creates a map of normalized module names to their slim module ids. So, unless the full name is used, the slim loader won't be able to figure out which module is being requested.

```javascript
// this is how the minimal loader "knows" what module to load during runtime
global.steal.map = {
  "myhub@1.0.0#puppies/puppies": 10,
  "myhub@1.0.0#weather/weather": 13
};
```

m-mujica:
Closes #759 

This way the slim user won't have to use full bundle names when dynamic importing stuff.

### DIFF
--- a/doc/guides/optimized_builds.md
+++ b/doc/guides/optimized_builds.md
@@ -8,7 +8,7 @@ as the default way of creating a build of a module and all of its dependencies; 
 
 Unlike regular [builds](steal-tools.build), optimized builds don't need to load or bundle StealJS at all; a thin wrapper is added instead to the main bundle so the browser can load and execute the modules correctly. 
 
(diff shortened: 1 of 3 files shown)


matthewp:
+1

Issue #740, opened 2017-06-13
See if we can switch to uglify-es


matthewp:
https://www.npmjs.com/package/uglify-es

This supports ES201x code. Assuming it's not breaking in some weird way with the current version of uglify I don't see why we couldn't switch to this in steal-tools 1.x.

jonlinnell:
Yes please – got some issues with uglify breaking when trying to process some ES6 code in a dependency, weirdly. Dev runs fine, but production build fails.

m-mujica:
@jonlinnell in case you missed it, we released steal-tools@1.6.0 with uglify-es, hopefully that fixes your production builds 💥

PR #786, opened 2017-07-12
Use UglifyJS with ES6+ support

### LINKED ISSUE #716: Suppress Uglify’s 32000 characters warning
In a new DoneJS app, I get this warning when I build my app:

```
WARN: Output exceeds 32000 characters
```

This is harmless but disturbing to see when building a new app. I think it’d be nice if we suppressed this warning. I think it’s something we could work around by specifying Uglify’s `max-line-len` option.

m-mujica:
Closes #716 and #740

success-gif

### DIFF
--- a/lib/build_types/minify_js.js
+++ b/lib/build_types/minify_js.js
@@ -18,23 +18,22 @@ exports.async = function(source, options) {
 };
 
 function uglify(source, options) {
+	var UglifyJS = require("uglify-es");
 	var envify = require("loose-envify/replace");
-	var UglifyJS = require("uglify-js");
 
 	var code = source.code;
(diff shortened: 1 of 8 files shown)


matthewp:
lgtm!

matthewp:
Shouldn't the require() call change too, since it's a different package?

lib/build_types/minify_js.js
```diff
@@ -18,23 +18,22 @@ exports.async = function(source, options) {
 };
 
 function uglify(source, options) {
+	var UglifyJS = require("uglify-es");
```
m-mujica: @matthewp do you mean this require?

matthewp: :disappear:

PR #787, opened 2017-07-14
Add options to target NodeJS for slim bundles

### LINKED ISSUE #719: Provide a way to target Node for slim bundles
The slim shim currently uses `<script>` to load bundles. This obviously won't work in Node, but that's ok.  I'm thinking the approach we take here is to have different targets. `browser` would be the default, we could create a `node` target (and possible a `worker` target in the future).

The Node target would substitute out the script loading code for something that uses Node APIs instead.

m-mujica:
Closes #719 

The docs for the new options are pending, will probably work on them once I finish working on the workers target.

@matthewp a couple of things

a) I changed the paths config for node so instead of being relative to the loader baseUrl, they are relative to the actual main file

```js
global.steal.paths = { "4": "main.js", "5": "foo.js" };
```
this is the only way I found to get the `require` to load bundles to work, thoughts?

b) The node slim loader still returns a promise when dynamically loading bundles, the reason being the transpiled dynamic imports expect a promise... We could go a step further and have transpile make those calls static, since bundles are loaded through `require`, not sure about this though.

c) when targets are provided, the buildResult objects are returned "labeled", the promise will resolve an object with this shape:

```js
{ web: { ...buildResult }, node: { ...buildResult } }
```
yay? nay?

### DIFF
--- a/lib/arrify.js
+++ b/lib/arrify.js
@@ -1,0 +1,14 @@
+var isArray = require("lodash/isArray");
+
+/**
+ * Converts a value to an array
+ * @param {*} value - The value to convert
+ * @return {Array}
+ */
+module.exports = function arrify(value) {
+	if (value == null) {
+		return [];
+	}
+
+	return isArray(value) ? value : [value];
+};
--- a/lib/build/slim.js
+++ b/lib/build/slim.js
@@ -1,27 +1,31 @@
 var pump = require("pump");
+var arrify = require("../arrify");
 var assign = require("lodash/assign");
+var makeDeferred = require("../make-deferred");
 var isUndefined = require("lodash/isUndefined");
 var assignDefaultOptions = require("../assign_default_options");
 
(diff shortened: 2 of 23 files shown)


matthewp:
a) that sounds more correct to me anyways, so no problem there for me.
b) a transpile sounds like overkill here so I think this is fine. the user was expecting a promise anyways and there could be some timing stuff that breaks if we make it synchronous.
c) looks good!

just answering the above questions, I still want to take a look at the code (which I haven't yet).

matthewp:
@m-mujica it looks like these changes didn't affect the web target's bundle sizes, is that correct? 👍 if so!

m-mujica:
@matthewp correct, I didn't have to change anything from the existing web loader, mostly remove stuff for the node version.

matthewp:
OK, can merge if you want.

PR #791, opened 2017-07-17
Document splitLoader and target

### LINKED ISSUE #785: [SLIM] Document `splitLoader` option
https://stealjs.com/docs/steal-tools.BuildOptions.html

m-mujica:
Closes #785 

screen shot 2017-07-17 at 08 45 09

### DIFF
--- a/doc/types/build-options.md
+++ b/doc/types/build-options.md
@@ -89,6 +89,53 @@ stealTools.build(config, {
 
 @option {Boolean} [watch=false] Actives watch mode which will continuously build as you develop your application.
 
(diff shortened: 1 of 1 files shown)


matthewp:
The docs should mention that `web` is the default target. Probably in the jsdoc way and a mention in the description text.

m-mujica:
@matthewp yeah, I thought about that.... The thing is, the default target is web but it's not the same things saying that `[target="web"]` b/c the output in that case would be written out to `dist/bundles/web/`, makes sense?

matthewp:
Ah, ok. Yeah, I think it's fine to ignore then, we treat it differently if target is not specified at all.

Issue #563, opened 2016-11-25
Minification breaks CJS dependency detection


m-mujica:
when my module is minified, the letter used in the CJS factory function for `require` is the same letter used later on in the same code to call a different function.

![screen shot 2016-11-25 at 16 33 12](https://cloud.githubusercontent.com/assets/724877/20634738/d1b29dce-b332-11e6-94c7-707d20671c7d.png)

![screen shot 2016-11-25 at 16 32 57](https://cloud.githubusercontent.com/assets/724877/20634743/d71cb4c0-b332-11e6-97b6-c117073744b3.png)

when I try to load the app locally, steal throws with a 404 loading "glob", which shouldn't be a dependency.

I'll fix this by preventing UglifyJS to mangle `require`

m-mujica:
I confirmed this by adding a build script to my app and setting the following option:

```javascript
stealTools.build(config, {
  mangle: {
    except: ['require']
  }
});
```

that fixed the problem 💥

m-mujica:
https://github.com/systemjs/systemjs/issues/1417#issuecomment-244943853

matthewp:
lodash is not hit with this bug (I think), so I think it should move up in priority.

matthewp:
Here's an example repo demonstrating the bug: https://github.com/DesignByOnyx/steal-lodash-bug

This is a particularly nasty one. I tried using `mangle: { reserved: ["require"] }` so that the `require` function is not renamed. However it still breaks because this code:

```js
var memoizeCapped = require('./_memoizeCapped');

/** Used to match property names within property paths. */
var reLeadingDot = /^\./,
    rePropName = /[^.[\]]+|\[(?:(-?\d+(?:\.\d+)?)|(["'])((?:(?!\2)[^\\]|\\.)*?)\2)\]|(?=(?:\.|\[\])(?:\.|\[\]|$))/g;
```

Gets reordered to look like:

```js
var reLeadingDot = /^\./,
    rePropName = /[^.[\]]+|\[(?:(-?\d+(?:\.\d+)?)|(["'])((?:(?!\2)[^\\]|\\.)*?)\2)\]|(?=(?:\.|\[\])(?:\.|\[\]|$))/g,
   memoizeCapped = require('./_memoizeCapped');
```

And because of the reordering, Steal's commonjs detection doesn't recognize the require() call as a dependency (because there is code to exclude `require` when inside of a string, and that regex breaks the string detection).

[...]

PR #792, opened 2017-07-20
Prevent uglify to reduce vars it breaks lodash build

m-mujica:
@matthewp 

Setting `compress.reduce_vars` to false "fixes" the build of the example app here 

https://github.com/stealjs/steal-tools/issues/563#issuecomment-316383978

but, it's a brittle solution, just moving around the `require` in the source files can trigger the issue again.... To replicate I had to copy/paste lodash module in my test. No idea how it decides when to put the require after the other variables.

James and Ryan tried to fix the regex but it's a hard problem when the code is minified, any unbalance quote before a require (in the same line) will break it. E.g

```js
var foo = `"`; bar = require("foo")`
```

or 

```js
var foo = /'/g; bar = require('foo')`
```

We could try to remove the es6 literals before detecting strings, but Regexes are more complicated because of the nested pairs `[]`, `()`. See https://stackoverflow.com/questions/172303/is-there-a-regular-expression-to-detect-a-valid-regular-expression and https://blogs.msdn.microsoft.com/jaredpar/2008/10/15/regular-expression-limitations/.

I tried beautifying the minified output using uglify's `beautify` option but that does not fix it, the beautifier has to be insert newlines in each var declaration so the bug does not appear.

Thoughts? (should I document these defaults before merging this PR if that's what we want to do?)

### DIFF
--- a/lib/assign_default_options.js
+++ b/lib/assign_default_options.js
@@ -10,7 +10,14 @@ module.exports = function(config, options){
 	options = assign({ // Defaults
 		minify: false,
 		bundleSteal: false,
-		uglifyOptions: {},
+		uglifyOptions: {
+			compress: {
+				reduce_vars: false
+			},
+			mangle: {
+				reserved: ["require"]
+			}
+		},
 		cleanCSSOptions: {
 			rebase: false,
 			inline: ["none"]
--- a/test/collapsed_vars/_memoizeCapped.js
+++ b/test/collapsed_vars/_memoizeCapped.js
@@ -1,0 +1,3 @@
+module.exports = function(cb) {
+	return cb();
+};
--- a/test/collapsed_vars/index.html
+++ b/test/collapsed_vars/index.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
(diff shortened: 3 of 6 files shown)


matthewp:
So is the regex approach to hard to fix? The only other option I can think of is to change transpile/amd to transpile to real AMD, like:

```js
define("name", ["dep1", "dep2"], function(dep1, dep2){

})
```

And then changing the require() calls to substitute in the argument variable. This would fix it definitively. (I would keep the cjs-wrapper AMD under a different format name, so that we can still use that for slim builds which don't have this issue).

m-mujica:
@matthewp I'm out of my depth with the regex, both Ryan/James tried some variations but I ran those regexes against a minified build and there were all sorts of mismatches. It'd take quite a while to get it right.

I like more the transpile approach, I'll give it a try tomorrow.

matthewp:
This would also work (I think), and would avoid the need to update all of the `require()` calls:

```js
define("name", ["require","exports", "module", "dep1", "dep2"], function(require, exports, module){

});
```

m-mujica:
@matthewp just tried that last one straight into the build and it does work, it just feels odd. I'm going to make the transpile changes.

PR #793, opened 2017-07-21
Make sure transpile does not break steal-tools

m-mujica:
Do not merge as is, just want to run the steal-tools tests with the updated transpile.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "steal-tools",
   "devDependencies": {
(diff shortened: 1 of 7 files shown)


m-mujica:
It seems an old version of steal installed by bower was causing some tests to fail. 

I still have some failures locally related to transform tests, as far as I can see the transform output is not happy with the `require` calls added in the transpiled output.

I need to do some research, not quite familiar with `transform` behaviour.

PR #801, opened 2017-07-26
Fix circular deps with Babel AMD's default import

m-mujica:
Closes https://github.com/stealjs/steal-tools/issues/802

- [x] Merge and release https://github.com/stealjs/transpile
- [x] Update package.json https://github.com/stealjs/steal-tools/pull/801/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R35

### DIFF
--- a/lib/graph/transpile.js
+++ b/lib/graph/transpile.js
@@ -1,5 +1,5 @@
-var _ = require("lodash");
-var transpile = require('transpile');
+var transpile = require("transpile");
+var cloneDeep = require("lodash/cloneDeep");
 var minify = require("../build_types/minify_js").sync;
 var processBabelPlugins = require("../process_babel_plugins");
 var processBabelPresets = require("../process_babel_presets");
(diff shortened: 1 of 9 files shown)


m-mujica:
What I came up with was to make steal-tools detect the circular dependency and then make transpile add a piece of code that makes sure there are no nested default properties.

```javascript
define("bar", ["exports", "./foo"], function(exports, _foo) {
  Object.defineProperty(exports, "__esModule", { value: true });

  var _foo2 = _interopRequireDefault(_foo);
  function _interopRequireDefault(obj) {
    return obj && obj.__esModule ? obj : { default: obj };
  }
  var name = (exports.name = "bar");
  var Bar = (function() {
    function Bar() {
      this.bar = _foo2.default.getBar();
    }
  })();
  exports.default = Bar;

  /*********************************************/
  /* code to fix the nested default reference */
  /********************************************/

  _interopCircularDependency(_foo2);
  function _interopCircularDependency(obj) {
    var __defaultExport;
    Object.defineProperty(obj.default, "default", {
      set: function(value) {
        if (obj.default.__esModule) {
          /* to be safe? */
          obj.default = value; /* fix the module reference */
        }
        __defaultExport = value;
     },
      get: function() {
        return __defaultExport;
      }
    });
  }
});
```

matthewp:
+:1

matthewp:
@m-mujica is this good to merge now?

m-mujica:
Yes, transpile@2.5.0 includes the latest changes.

PR #805, opened 2017-08-04
Update the slim guide with cordova/electron build steps

m-mujica:
I think we can close 

https://github.com/stealjs/steal-cordova/issues/16 and
https://github.com/stealjs/steal-electron/issues/22

when this is merged.

screen shot 2017-08-04 at 16 23 53

screen shot 2017-08-04 at 16 24 03


matthewp:
👍

PR #808, opened 2017-08-15
Make 'target' property optional for optimize CLI

### LINKED ISSUE #809: optimize cli: target property should be optional
running `steal-tools optimize` will cause yargs to throw with:

```
Invalid values:
   Argument: target, Given: undefined, Choices: "web", "node", "worker"
```

m-mujica:
Closes #809

Unfortunately there is yargs bug preventing me to use their API to make the array as optional, my ad-hoc solution was to copy their description including the possible values which is nice

screen shot 2017-08-14 at 18 23 49

But removed the `choices` property, the `optimize` method already validates `target` values. I put a test there just to make sure it works as intended.

### DIFF
--- a/lib/cli/cmd_optimize.js
+++ b/lib/cli/cmd_optimize.js
@@ -26,8 +26,11 @@ var options = assign(
 		},
 		target: {
 			type: "array",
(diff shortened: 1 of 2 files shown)


matthewp:
👍

PR #813, opened 2017-08-15
Make sure Steal does not break the build

m-mujica:
teal@1.5.7 introduced an issue affecting side bundles and builds where steal is bundled with main


m-mujica:
I'm going to merge this one so I can rebase my other PR.

PR #824, opened 2017-08-24
buildType flag

pYr0x:
adds a flag to the buildResult to identify which build process was running.


matthewp:
Awesome, thank you!

PR #859, opened 2017-09-29
Stop plucking @loader from graph in optimized builds

### LINKED ISSUE #856: @loader not defined in slim build
this PR and this line cause an error after build
https://github.com/stealjs/steal-tools/pull/850/files#diff-970f87aa3e6b874cd244518372a6e3f5R90
`moduleId` is `@loader` which is not defined.
the transpiled code looks like this:
```[
		151,
			var atLoader = stealRequire("@loader");
			var steal = {};
			steal.loader = atLoader;
			steal.done = function() {
				return Promise.resolve();
			};
			stealModule.exports = steal;
		}
	],
```

m-mujica:
Closes #856

### DIFF
--- a/lib/build/slim.js
+++ b/lib/build/slim.js
@@ -64,7 +64,10 @@ module.exports = function(cfg, opts) {
 		addModuleIds(),
 		convertSlimConfig(),
 		loadNodeBuilder(),
-		transpile({ outputFormat: "slim" }),
+		transpile({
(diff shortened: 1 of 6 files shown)


m-mujica:
cc: @pYr0x this was the actual issue with `@loader`.

pYr0x:
Thanks

PR #864, opened 2017-10-01
Improve minified code sourcemaps

### LINKED ISSUE #863: UglifyJS and sourcemaps
- [x] sourcesContent is only included if there was an existing map
- [x] SourceMaps were only generated when there was an existing map regardless of whether 
       `BuildOptions.sourceMaps` value

m-mujica:
Closes #863


matthewp:
Thanks!

PR #870, opened 2017-10-04
 Set environment global as context of module factory

### LINKED ISSUE #868: Slim output breaks `this`
Some bundlers out there use the wrapper below for universal module definitions:

```js
(function(root, factory) {
})(this, function() {
  var foo = { bar: "bar" };
   return foo;
});
```

`this` is expected to be bound to the `window` in browsers, the slim loader calls each module factory function binding `this` to `stealExports` breaking some code.

What I have in mind is to detect the ThisExpression in the AST and wrap the UMD in a IIFE so `this` is bound to window.

```js
(function() {
  (function(root, factory) {})(this, function() {
    var foo = { bar: "bar" };
    return foo;
  });
})();
```

The tricky part here would be detect the ThisExpression bound to the window. Hopefully I can get some scope information from the AST.

m-mujica:
Merge https://github.com/stealjs/steal-tools/pull/866 first 

Closes #868

### DIFF
--- a/lib/node/slim/make_shim_template.js
+++ b/lib/node/slim/make_shim_template.js
@@ -95,6 +95,15 @@ var renderMainImportPartial = function(options) {
 	return result;
 };
 
+/**
+ * Returns the name of the global based on the build target
+ * @param {string} target - The target build name
+ * @return {string} defaults to "window" if target is falsy.
+ */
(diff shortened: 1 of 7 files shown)


matthewp:
LGTM once tests pass

PR #896, opened 2017-11-28
add npmignore

pYr0x:
same as on stealjs, ignore all unnecessary folder/files for npm


pYr0x:
@matthewp the test break on node 8. this PR does not affect the tests. so merging without fear :)

PR #922, opened 2018-01-11
Test steal-bundler does not move assets below dest

### LINKED ISSUE #882: `bundleAssets` does not put assets in specified dest folder
Whenever specifying `dest` option, the assets are not put in the destination, but rather in the destination's parent. With the default `dest` value, you end up with this:

image

... but if you set `dest: 'public/noofoo'`, you end up with this:

image

Note that the generated `public/node_modules` folder is not located inside the dest folder. This isn't blocking us at the moment, but it means that we have to manage 2 dist folders: `public/dist` and `public/node_modules` (unless we want to use the default "dest" option... which we don't).

It's worth mentioning that the reason this not blocking us is because we are building to the default location and then copying to our desired destination.

m-mujica:
Closes #882


package.json
```diff
@@ -26,7 +26,8 @@
     "pdenodeify": "^0.1.0",
     "prettier": "1.9.2",
     "pump": "^1.0.2",
-    "steal-bundler": "^0.3.5",
+    "steal": "^1.6.4",
+    "steal-bundler": "stealjs/steal-bundler#dest",
```
m-mujica: I'm going to update this once steal-bundler PR is merged and released.

PR #932, opened 2018-01-22
Allow plugins to force inclusion of the slim dynamic loader 

### LINKED ISSUE #931: Missing stealRequire.dynamic with done-autorender
A simple app with a main like:

```html
<html>
	<head>
		<title>Hello page</title>
	</head>
	<body>
		<can-import from="test/basics-optimized/state" export-as="viewModel"/>
		<h1 id="hello">Hello {{hello()}}!</h1>
	</body>
</html>
```

Fails to load using the slim build, since `done-autorender` will dynamically load `test/basics-optimized/state` but the loader won't include the progressive loading stuff because there are not secondary bundles written out (`test/basics-optimized/state` is included in the main bundle)

### Solution

My solution (which I'm not happy with) is to output the progressive loading pieces of the slim loader when there are secondary bundles written out OR done-autorender is in the graph.

Ideally we should be able to detect that there are dynamic imports in the codebase and extend the loader accordingly but that detection is made during transpile; unless we use a flag in a load object or something, how does steal-tools know about the dynamic loading?

Also, in the regular build if dynamic loading is used, there will be some config needed to create new bundles, done-autorender is just a strange use case in that regard.

m-mujica:
Closes #931


m-mujica:
@matthewp thoughts about this?

matthewp:
👍 lgtm

PR #935, opened 2018-01-23
Slim should use plugins with dynamic loading too

### LINKED ISSUE #934: Optimized loader loads css bundles with a script tag
If there is a bundle that requires a plugin to be used (css) and that's dynamically loaded, `stealRequire.dynamic` tries to load it as a regular javascript bundle causing weird errors.

screen shot 2018-01-23 at 11 44 00 am

m-mujica:
Closes #934


matthewp:
👍

PR #938, opened 2018-01-30
Add shared bundles support to the slim loader

### LINKED ISSUE #936: Slim loader does not load shared bundles
Only the "master" bundle is loading, causing the loader to throw trying to execute module that are in shared bundles.

m-mujica:
Closes #936


matthewp:
👍 lgtm

PR #940, opened 2018-01-30
Do not use a single normalizeMap for all bundles

### LINKED ISSUE #939: incorrect bundle path with steal-tools@1.11.1
I ran into a bug running the canjs test suite with steal-tools@1.11.1. For some reason the path to `can.js` does not contain `/global` with the most recent patch. You can reproduce this by checkout out [canjs#master](https://github.com/canjs/canjs/), running `npm run build && npm run build-tests`, and then loading the `test/production.html` file in the browser.

With steal-tools@1.11.0:
image

With steal-tools@1.11.1:
image

m-mujica:
Closes #939 

Before, We made both `normalizeMap` and `slimConfig` be a unique object reference for all bundles, but `normalizeMap` is used during transpilation and having a single references causes some module to be transpiled incorrectly because they are inhering state from previous bundles.

This change might affect slim bundles using `steal-conditional` in different bundles (although I don't think it matters since `normalizeMap` would be used right away by transpile and there is no need to collect everything like `slimConfig` for example)

I'm going to open a new issue to write a test for the slim build and make sure it works, but this fixes the issue affecting the CanJS build.

cc: @matthewp

### DIFF
--- a/lib/graph/make_graph.js
+++ b/lib/graph/make_graph.js
@@ -15,7 +15,7 @@ var formatMap = {
 
 var SystemRegistry = global.SystemRegistry = {};
 
-module.exports = function makeGraph(config, options) { 
+module.exports = function makeGraph(config, options) {
 	options = options || {};
 	winston.info('OPENING: ' + (config.main || config.config || config.configMain));
 	var oldGlobalSystem = global.System;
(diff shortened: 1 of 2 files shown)


m-mujica:
Ref: https://github.com/canjs/canjs/pull/3909

matthewp:
lgtm

PR #942, opened 2018-01-31
Make BuildOptions.target use stealTools.optimize

### LINKED ISSUE #941: BuildOptions target docs example uses stealTools.build
But this option is only available for optimized builds:

https://stealjs.com/docs/steal-tools.BuildOptions.html

screen shot 2018-01-31 at 10 09 17 am

m-mujica:
Closes #941

### DIFF
--- a/doc/types/build-options.md
+++ b/doc/types/build-options.md
@@ -127,7 +127,7 @@ like any other bundle.
 Usage:
 
 ```js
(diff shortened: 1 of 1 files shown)


m-mujica:
I'm going to merge right away, the build shouldn't be an issue.

PR #950, opened 2018-02-22
Allow spaces in special comments

m-mujica:
//! steal-remove-start
//! steal-remove-end

That works now.

Closes https://github.com/stealjs/steal/issues/1329

### DIFF
--- a/lib/build/clean.js
+++ b/lib/build/clean.js
@@ -1,14 +1,24 @@
-module.exports = function(original, options) {
+function clean(original, options) {
 	var result = original;
 	var removeTags = options.removeTags || [];
 
-	if(!options.dev) {
-		removeTags.push('steal-remove');
+	if (!options.dev) {
+		removeTags.push("steal-remove");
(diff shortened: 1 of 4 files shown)


matthewp:
lgtm

roemhildtg:
Found another issue @m-mujica 

In windows, often times newlines aren't using the `\n` character (I wish they were always but occasionally they're not) and this can lead to hard to figure out issues...as development code isn't getting removed in production etc...

the solution: also check for `\r` character?
```

function makeTagRegEx(tag) {
	return new RegExp(
		"(\\s?)//!(\\s?)" + tag + "-start((.|\n|\r)*?)//!(\\s?)" + tag + "-end",
		"gim"
	);
}
```

chasenlehara:
@roemhildtg Would you mind opening a new issue for that please?

roemhildtg:
Sure @chasenlehara sorry I was going to do that, and got sidetracked...

PR #965, opened 2018-03-29
Envify should work regardless of minify setting

### LINKED ISSUE #925: Disabling minify breaks envify option
Using `minify: false` with `envify: true` breaks the `envify` option.

```

var buildPromise = stealTools.build({
}, {
  envify: true,
  minify: false
});
```

m-mujica:
Closes #925

Envify was applied as part of the minification logic, but the two things are independent so I made it its own stream (probably slows down build time b/c we are walking the graph one more time)


matthewp:
giphy

Issue #974, opened 2018-04-02
Add Tree Shaking


matthewp:
This issue spins off of the [tree shaking RFC](https://github.com/stealjs/rfcs/pull/20).

With regard to the API tree-shaking will be on by default. It can be disabled in BuildOptions:

```js
stealTools.build({}, {
  treeShaking: false
});
```

Also we will following the `sideEffects` value in package.json, like webpack does. This means for 3rd party packages we will only tree shake those modules if they have `"sideEffects": false` in their package.json.

This can be override by passing the `treeShakingForce` flag:

```js
stealTools.build({}, {
  treeShakingForce: true
});
```

We will emit a warning when force is turned on.

----

Implementation wise, tree shaking will be a separate step in the process that will happen before transpile. This is because we need the untranspiled source to know what the imports and exports are for each module.

At first tree-shaking will only happen from ES source modules. Adding CommonJS will occur time-permitting and should following some safety rules like 
* Only tree-shake if the CJS module only modules `module.exports` and/or `exports` in the module body.
* Do not tree-shake if the CJS module assigns itself to some local variable like `var foo = module.exports` because that would then mean we need to verify that `foo` is never modified.

--- 

Documentation to add:

* [ ] - Add a topic page for Tree Shaking under the __Bundling__ section.
* [ ] - Document `treeShaking` and `treeShakingForce` options under BuildOptions.
* [ ] - Document the `sideEffects` package.json boolean.

PR #984, opened 2018-04-20
Implement tree shaking

matthewp:
This pull request implements tree shaking. It uses rollup under the hood for the heavy lifting so we don't have to worry about maintaining complex dead-code elimination logic. This will be released as a prerelease first to ensure existing applications are not negatively effected.

It is easily turned off with the `--no-tree-shaking` cli flag of `treeShaking: false` from the JavaScript API.


matthewp:
Closes #974

PR #1002, opened 2018-05-24
Do not add nodes to graph after bundles property is added

### LINKED ISSUE #1001: Cannot create development bundles for myhub
While creating dev bundles following the guide, I got an error with `node.bundles` being undefined.

m-mujica:
screen shot 2018-05-24 at 3 25 34 pm

Closes #1001


m-mujica:
Not sure what happened to coveralls, but it should be ok since I added more tests.

PR #1003, opened 2018-05-26
Prevent npm preload module to break dev bundles

### LINKED ISSUE #947: Minified Dep Bundles are Broken
When building the CanJS site if I use `"deps-bundle": "steal-tools bundle --deps"` in my scripts and try to load a demo with that deps-bundle like:
```
<script id="demo-source" main="@empty" src="../../node_modules/steal/steal.js" deps-bundle>
```
It will break on load with the error:
```
steal.js:7390 Error: Error loading "dev-bundle" at http://localhost:8080/dev-bundle.js
TypeError: define is not a function
	    at Object.eval (http://localhost:8080/dev-bundle.js:4:125077)
	    at eval (http://localhost:8080/dev-bundle.js:5:4)
	    at eval (http://localhost:8080/dev-bundle.js:5:98)
	Evaluating http://localhost:8080/dev-bundle.js
```

If I add the `--no-minify` option `"deps-bundle": "steal-tools bundle --deps --no-minify"` it works fine but the deps-bundle is not minified.

m-mujica:
Closes #947 

Part of the problem is that the module to preload the npm packages is added at the top of the bundle. So we run into situations like:

```js
{ foo: "/*" };define("foo", function() { var bar = "*/" });
```

then the amd extension can't see the "define" because the comments regex matches a bunch of the module definition:

screen shot 2018-05-26 at 11 05 36 am

The solution I came up with was to push the module with the package.jsons to the bottom of the bundle, that way the first thing the regex matches is the actual AMD modules.

Placing the preloader at the bottom shouldn't matter since it gets executed right away.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ test/transform_export/out.js
 test/exports_basics/out.js
 test/circular/export/
 test/serviceworker/service-worker.js
+!test/dev_bundles_minify/node_modules
--- a/lib/bundle/add_npm_packages_node.js
+++ b/lib/bundle/add_npm_packages_node.js
@@ -1,3 +1,4 @@
+var prettier = require("prettier");
 
 /**
  * Adds node to bundle to preload npm packages for dev bundles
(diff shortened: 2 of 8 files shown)


m-mujica:
screen shot 2018-05-26 at 6 57 19 am

matthewp:
Sounds good to me, nice job.

PR #1010, opened 2018-06-01
Adds a test for building steal project on windows

matthewp:


matthewp:
Currently breaking on windows. Will re-run when new steal-less is released.

PR #1036, opened 2018-07-16
provide glob pattern for bundles

pYr0x:
in steal you can add bundles that are loaded progressive 
see this HowTo: https://stealjs.com/docs/StealJS.guides.progressive_loading.html#build-a-progressive-loading-production-app

this adds feature for #507 and new version of #512 
this is available in steal version 2.0

@matthewp this version is similar to #512. if we are with npm then it removes the package name, find the files, and adds back the package name. https://github.com/stealjs/steal-tools/blob/glob-bundle/lib/loader/find_bundle.js#L75
weird mapping is not supported. just simple globs like
* `bundle: ["npm-bundle/sites/**/*"]`
* `bundle: ["npm-bundle/components/**/*","!npm-bundle/components/**/*-test.js"]`

this feature is fully backwards compatible

hopefully this lands on steal-tools. this save me a lot of writing bundle names

### DIFF
--- a/lib/loader/find_bundle.js
+++ b/lib/loader/find_bundle.js
@@ -1,3 +1,4 @@
+var path = require("path");
 var glob = require("globby").sync;
 
 module.exports = findBundles;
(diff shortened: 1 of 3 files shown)


matthewp:
Awesome! Will definitely get this merged in.

matthewp:
@pYr0x this needs to be documented, right?

pYr0x:
Yep. I can write some docs. 
Under https://stealjs.com/docs/steal-tools.StealConfig.html ?

matthewp:
Wouldn't it be here? https://stealjs.com/docs/config.bundle.html . Looks like there's already a glob example though. Did yours just improve the existing support?

pYr0x:
globbing was only available if a plain string was provided. like the example shown.

i will update the docs and refer a link to [StealConfig]( https://stealjs.com/docs/steal-tools.StealConfig.html) on https://stealjs.com/docs/config.bundle.html

PR #1047, opened 2018-07-22
substring different on windows

### LINKED ISSUE #1046: process.cwd substring modules falsely on windows
i had the same problem as @roemhildtg
his comments on gitter: https://gitter.im/stealjs/steal?at=5b536af0ee530e4aac8b471c

it seems we are the only one using windows :)

the bug is on https://github.com/stealjs/steal-tools/blob/master/lib/graph/treeshake.js#L222
the problem is, that on macOS the path on `process.cwd()` starts with `/`. on windows `process.cwd()` starts with `C:/`

so we need to check the os and substring different https://github.com/stealjs/steal-tools/blob/master/lib/graph/treeshake.js#L19 if we are on windows

pYr0x:
check OS and substring different on windows
close #1046

### DIFF
--- a/lib/graph/treeshake.js
+++ b/lib/graph/treeshake.js
@@ -1,5 +1,6 @@
 "use strict";
 
+var os = require("os");
 var babel = require("babel-standalone");
 var entries = require("object.entries");
 var dependencyResolver = require("../node/dependency_resolver");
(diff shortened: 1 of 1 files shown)


lib/graph/treeshake.js
```diff
@@ -17,6 +18,9 @@ function treeshake(data) {
 	let getNode = id => data.graph[id];
 	let mains = uniq(data.mains.concat(data.loader.bundle));
 	let cwd = process.cwd().substr(1);
+	if (os.type().toLowerCase() === "windows_nt"){
```
matthewp: Is this the best way to check if it's windows? In the past I've used `process.platform === 'win32'`.

pYr0x: hmm i dont know. docs said `Windows_NT` https://nodejs.org/api/os.html#os_os_type

stackoverflow said  `process.platform` https://stackoverflow.com/questions/8683895/how-do-i-determine-the-current-operating-system-with-node-js

matthewp: Ok, I'd think either will work then.

Issue #1050, opened 2018-07-24
import "side-effect" gets tree-shaken if the module is not ES


pYr0x:
here is a demo app.
https://github.com/pYr0x/css-treeshaken

using steal and steal-tools -> 2

the options `treeShaking: false` addes the css files as dependencies to the "main" module

matthewp:
Looks like it's building out, just removed as a dependency. Looking into it.

matthewp:
Breaking test: https://github.com/stealjs/steal-tools/commit/1c8c19253908660e6e809ce1cd40ecf02e68b6ff

The problem is not related to CSS, just doesn't work if the dependency is not an ES module.

PR #1052, opened 2018-07-24
Adds a breaking test for import a module for side effects

matthewp:
This shows that importing a module for side effects doesn't work if that
module is not ES.


matthewp:
Fixes #1050

PR #1080, opened 2018-10-04
Update package.json

frank-dspeed:
using mozilla source map as bitovi has not done any changes and is outdated i think the fork can be deprecated at all


matthewp:
The fork worked around a bug in uglify. I would like to switch to the non-fork version but would need to verify that the problem no longer exists. Anyways, this PR fails CI.

PR #1114, opened 2019-03-05
Add option to keep ES2015 code in built code

m-mujica:
Instead of always forcing code to be transpiled to ES5

*DO NOT MERGE* ... we should wait until transpile and steal are released.


matthewp:
Once deps are updated 👍

PR #1145, opened 2019-10-26
Fix build of ES2015 code with forceES5 set to false

### LINKED ISSUE #1144: Cannot build ES2015 without ES5 transpilation
```js
let name = "John";
let time = "today";

export default `Hello ${name}, how are you ${time}?`
```

steal-tools throws an error with this code if `forceES5` is set to false.

> Setting `sourceMaps: false` gets the build working

I tracked down the problem to the traspile package which uses a very old version of `estraverse` when trying to generate sourcemaps.

m-mujica:
Closes #1144

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "through2": "^3.0.0",
     "tmp": "0.0.33",
     "traceur": "0.0.111",
-    "transpile": "^2.7.0",
+    "transpile": "^2.7.2",
     "urix": "^0.1.0",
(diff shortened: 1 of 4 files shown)


m-mujica:
@matthewp any suggestions to debug https://ci.appveyor.com/project/matthewp/steal-tools-427#L189 ? 

It works on my machine™

matthewp:
Node version maybe?

m-mujica:
I'm going to merge and release this.

PR #1148, opened 2019-11-07
Fix appveyor build

m-mujica:
- A treeshaking test was using a symlink file (css.js) which does
  not work correctly when tests are running on windows. This change
  replaces the link with the actual file.

- Upgrade testee to latest version

- Remove wilcards from jshint script
  The script does not work in Windows when wildcards are used:
    `ERROR: Can't open lib/**/*.js (seen on appveyor build)`
  jshint recursively lints all files when a driectory is passed to
  the CLI so we can pass "lib/" instead of "lib/**/*.js".

- Use latest 8.x version available

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@
 environment:
   matrix:
   # node.js
-  - nodejs_version: "8.0.0"
+  - nodejs_version: "8"
 
 # Install scripts. (runs after repo cloning)
 install:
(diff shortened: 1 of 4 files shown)


m-mujica:
Hmmm the bulid now gets stuck and I have no idea why. I'll take a look at this later.

https://ci.appveyor.com/project/matthewp/steal-tools-427/builds/28696315#L635

m-mujica:
tenor

matthewp:
👍

PR #1149, opened 2019-11-17
Fix race condition where last format transform overrides previous ones

### LINKED ISSUE #1139: Build exports wrong cjs format
Currently the export for `cjs` is the same as `amd` I noticed this in 2 `canjs` packages, `can-log` and `can-key-tree` .

The used `steal-tools` version `"steal-tools": "^2.2.1"`

can-log_js_—_can-log

m-mujica:
The transform function resolved by `transformInput` was handling
the bundle concatenation to the event loop too early, causing future
transforms (sharing the same graph) to reset the activeSourceKeys
object -where the result of transpiled graphs are kept- before the bundle
was concatenated.

This meant the last transform would override all of the previous ones,
instead of doing `return Promise.resolve().then(concatenateBundle)`, this
commit creates a new promise that runs `concatenateBundle` right away
instead of waiting for the next tick (and opening the window from future
transforms to mutate the graph).

Closes #1139

### DIFF
--- a/lib/build/transform.js
+++ b/lib/build/transform.js
@@ -143,31 +143,33 @@ var transformImport = function(config, transformOptions){
 				winston.debug("+ %s", node.load.name);
 			});
 
-			return Promise.resolve()
-			.then(function(){
+			return new Promise(function(resolve, reject) {
 				var concatOptions = {
 					excludePlugins: options.format === "global",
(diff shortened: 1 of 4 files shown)


m-mujica:
tenor

matthewp:
lgtm, awesome work

PR #1153, opened 2019-12-25
Update Node version to 13

cherifGsoul:
Closes #1151 

Update node engine field in `package.json` and `.travis.yml` to `13.x`.

PR #1157, opened 2020-06-06
Add support for node 14 (modelled after #1153)

icarito:
Untested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
   - 8
-  - 13.5.0
+  - 14.4.0
 before_script:
   - npm dedupe
 sudo: false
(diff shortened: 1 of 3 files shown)


icarito:
This works in my scenario, but there are a few warnings.
```
(node:1556) Warning: Accessing non-existent property 'padLevels' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:2236) Warning: Accessing non-existent property 'VERSION' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
```
Also a lot of warnings about outdated dependencies (e.g. Chokidar, etc, see AppVeyor).
I guess those are separate issues.

matthewp:
Thanks!

